PR #49, opened 2017-11-21
Gave code block comment a break in README

BenEmdon:
Why?
 
nov-20-2017 14-45-55

View the [rich changes ](https://github.com/BenEmdon/FunctionalTableData/blob/ba27437f11898f58f08574be4cecb75743d33cd3/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,8 @@ let section = TableSection(key: "header-unique-key", rows: [])
 Each section contains a series of rows where each row value must conform to the `CellConfigType` protocol.
 
 ```swift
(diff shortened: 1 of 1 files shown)


raulriera:
> Why?

Because text editor can wrap these if you wish to or not. Obviously this is not the case for github markdown rendering

BenEmdon:
> Because text editors can wrap these if you wish to or not. Obviously this is not the case for github markdown rendering

I wasn't trying to throw shade 💅  at you @raulriera I was explaining why I made the change with a rhetorical question
sorry if that came off as shade

raulriera:
I thought you were just asking :)

Issue #37, opened 2017-11-01
Expand CellStyle to allow cornerRadius


raulriera:
## What

Implant a way to clip the contents of a cell. Specially useful for UICollectionViewCells. Maybe adding a mask property to CellConfig?

Issue #32, opened 2017-10-19
Port the changes to allow separator colours


raulriera:
## What

Port this code over ios/pull/1311/commits/5e3b30fd6721459c5593dc1b190d35ef23eca9af

raulriera:
Closed by https://github.com/Shopify/FunctionalTableData/pull/59

PR #59, opened 2017-12-12
Support for separator colour changes, collection view clipping and more

raulriera:
## What

Fixes https://github.com/Shopify/FunctionalTableData/issues/37 and https://github.com/Shopify/FunctionalTableData/issues/32

### DIFF
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -138,6 +138,7 @@ extension CellStyle: Equatable {
 		equality = equality && lhs.backgroundView == rhs.backgroundView
 		equality = equality && lhs.tintColor == rhs.tintColor
 		equality = equality && lhs.layoutMargins == rhs.layoutMargins
+		equality = equality && lhs.cornerRadius == rhs.cornerRadius
 		return equality
 	}
 }
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -13,6 +13,7 @@ public struct CellStyle {
 	public static var selectionColor: UIColor? = nil // TODO: make this not a static like this
 	public var bottomSeparator: Separator.Style?
 	public var topSeparator: Separator.Style?
+	public var separatorColor: UIColor?
 	public var highlight: Bool?
 	public var accessoryType: UITableViewCellAccessoryType = .none
 	public var selectionColor: UIColor? = CellStyle.selectionColor
@@ -20,18 +21,22 @@ public struct CellStyle {
 	public var backgroundView: UIView?
 	public var tintColor: UIColor?
 	public var layoutMargins: UIEdgeInsets?
+	public var cornerRadius: CGFloat
 	
 	public init(topSeparator: Separator.Style? = nil,
 	            bottomSeparator: Separator.Style? = nil,
+	            separatorColor: UIColor? = nil,
 	            highlight: Bool? = nil,
 	            accessoryType: UITableViewCellAccessoryType = .none,
 	            selectionColor: UIColor? = CellStyle.selectionColor,
 	            backgroundColor: UIColor? = nil,
 	            backgroundView: UIView? = nil,
 	            tintColor: UIColor? = nil,
-	            layoutMargins: UIEdgeInsets? = nil) {
+	            layoutMargins: UIEdgeInsets? = nil,
+	            cornerRadius: CGFloat = 0) {
 		self.bottomSeparator = bottomSeparator
 		self.topSeparator = topSeparator
+		self.separatorColor = separatorColor
 		self.highlight = highlight
 		self.accessoryType = accessoryType
 		self.selectionColor = selectionColor
@@ -39,6 +44,7 @@ public struct CellStyle {
 		self.backgroundView = backgroundView
 		self.tintColor = tintColor
 		self.layoutMargins = layoutMargins
+		self.cornerRadius = cornerRadius
 	}
 	
 	func configure(cell: UICollectionViewCell, in collectionView: UICollectionView) {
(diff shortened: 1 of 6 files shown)


FunctionalTableData/TableView/TableCell.swift
```diff
@@ -28,6 +28,33 @@ public class TableCell<ViewType: UIView, Layout: TableItemLayout>: UITableViewCe
 		prepare?(view)
 		prepare = nil
 	}
+	
+	public override func accessibilityElementCount() -> Int {
```
raulriera: How come this is new? 🤔 I don't remember what we did to this

FunctionalTableData/CellStyle.swift
```diff
@@ -116,6 +130,7 @@ extension CellStyle: Equatable {
 	public static func ==(lhs: CellStyle, rhs: CellStyle) -> Bool {
 		var equality = lhs.bottomSeparator == rhs.bottomSeparator
 		equality = equality && lhs.topSeparator == rhs.topSeparator
+		equality = equality && lhs.separatorColor == rhs.separatorColor
 		equality = equality && lhs.highlight == rhs.highlight
 		equality = equality && lhs.accessoryType == rhs.accessoryType
 		equality = equality && lhs.selectionColor == rhs.selectionColor
```
Caiopia: Should we add `cornerRadius` here too?

FunctionalTableData/TableView/TableCell.swift
alunbestor: I think we deliberately omitted this fuckery when we ported stuff back to the FTD repo last time. Let's leave it in otherwise we'll go insane.

FunctionalTableData/CellStyle.swift
```diff
@@ -57,24 +63,32 @@ public struct CellStyle {
...
 		if let selectionColor = selectionColor {
 			let selectedBackgroundView = UIView()
 			selectedBackgroundView.backgroundColor = selectionColor
 			cell.selectedBackgroundView = selectedBackgroundView
 		} else {
 			cell.selectedBackgroundView = nil
 		}
+		
+		cell.layer.cornerRadius = cornerRadius
+		cell.layer.masksToBounds = true
```
alunbestor: I wonder if `masksToBounds` should only be set if `cornerRadius` is > 0. We don't need this otherwise, and theoretically it makes drawing slower; also, consumers of this API may want to display content that extends beyond the bounds of the cell.

FunctionalTableData/CellStyle.swift
```diff
@@ -88,14 +102,14 @@ public struct CellStyle {
...
-        
-        cell.contentView.layoutMargins = layoutMargins ?? UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
+		// SUPER HACK! On iOS 11, setting preserveSuperviewLayoutMargin to true changes the behavior
+		// of the layout margins, even when it was already true. Without this fix our layout margins
+		// were not being respected and our cells were shorter than on iOS 10
+		if cell.contentView.preservesSuperviewLayoutMargins {
+			cell.contentView.preservesSuperviewLayoutMargins = true
+		}
+		
+		cell.contentView.layoutMargins = layoutMargins ?? UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
```
alunbestor: This insets value ought to be a class constant somewhere.

FunctionalTableData/CellStyle.swift
```diff
@@ -116,13 +130,15 @@ extension CellStyle: Equatable {
 	public static func ==(lhs: CellStyle, rhs: CellStyle) -> Bool {
 		var equality = lhs.bottomSeparator == rhs.bottomSeparator
 		equality = equality && lhs.topSeparator == rhs.topSeparator
+		equality = equality && lhs.separatorColor == rhs.separatorColor
 		equality = equality && lhs.highlight == rhs.highlight
 		equality = equality && lhs.accessoryType == rhs.accessoryType
 		equality = equality && lhs.selectionColor == rhs.selectionColor
 		equality = equality && lhs.backgroundColor == rhs.backgroundColor
 		equality = equality && lhs.backgroundView == rhs.backgroundView
```
alunbestor: This view comparison should probably use `isEqual` instead. AFAIK that defaults to pointer equality for `UIView`s anyway, but consumers of the API may want to avoid an unnecessary style update for custom views they know are functionally identical.

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -181,20 +181,37 @@ public class FunctionalCollectionData: NSObject {
...
-			}
-			
-			for section in newSections {
-				if Set(section.map { $0.key }).count != section.rows.count {
-					let rowKeys = section.rows.map { $0.key }.joined(separator: ", ")
-					let reason = "\(strongSelf.name) : Section.Row keys must all be unique"
-					let userInfo: [String: Any] = ["Section": section.key, "Duplicates": rowKeys]
+			func validateKeyUniqueness() {
+				let sectionKeys = newSections.map { $0.key }
+				if Set(sectionKeys).count != newSections.count {
```
alunbestor: Conditions like this should be `guard`s.

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -212,9 +229,8 @@ public class FunctionalCollectionData: NSObject {
 		
 		let oldSections = sections
 		
-		var visibleIndexPaths: [IndexPath] = []
-		DispatchQueue.main.sync {
-			visibleIndexPaths = collectionView.indexPathsForVisibleItems.filter {
+		let visibleIndexPaths = DispatchQueue.main.sync {
```
alunbestor: 😚👌

FunctionalTableData/CellStyle.swift
g-Off: == should be doing isEqual for NSObjects, === does pointer comparisons

alunbestor: TIL! 😮

PR #62, opened 2017-12-13
Allow these to be read outside the module

raulriera:
## What

The exception handlers sometimes check these properties, allow them to be used outside of the module


raulriera:
Oops, didn't mean to close this so far

PR #64, opened 2018-01-05
add support for missing UIScrollViewDelegate functions

robinsenior:
I was making a test app and one of the UIScrollViewDelegate functions I needed wasn't propagated by FunctionalTableData, so I added it and added the other missing functions to to FunctionalTableData and FunctionalCollectionData while I was at it.


g-Off:
There are potential performance implications of hooking up all the delegate methods. IIRC `UIScrollView` caches which delegate methods its delegate implements when its delegate is set and when events happen it doesn't even attempt to fire those calls. By hooking up to all of them we're going to be triggered for everything, even if we don't care about that particular delegate method

robinsenior:
@g-Off ah, interesting. let me look around and see if I can find any metrics on that.

Issue #54, opened 2017-11-28
Implement a reusable case


TheoBdev:
## What 

Right now, `updateView` is not clear enough about reusability. 

## Suggested approach

@robinsenior mentioned http://khanlou.com/2017/03/that-one-optional-property/ : 

```
enum UseCase {
    case prepareForReuse
    case update(LabelState)
}
```

BenEmdon:
I think that's a great Idea and would prevent people from forgetting to prepare for reuse 👏

PR #65, opened 2018-01-05
proposal for cell reuse refactor

robinsenior:
Re: https://github.com/Shopify/FunctionalTableData/issues/54

If we want to go ahead with this it'll require a lot of refactoring in our internal projects. Thoughts?


raulriera:
We shouldn't introduce a major change like this without support for the previous implementation. Although this is safer than the `State?` I think this approach adds too much noise to the implementations given that everything needs to `switch` on the state

robinsenior:
@raulriera alternatively we could add a `cellPrepareForReuser` (🙃)  alongside the existing `cellUpdater`

TheoBdev:
I love the change @robinsenior ! Thanks for taking care of that! 
@raulriera : imo it looks much clearer from an implementation stand point if you were to use FTD for the first time

raulriera:
I nevera saw these replies, it certainly looks ea sidra to read but way more typing in my opinion. i like it, but it should be a companion API and deprecate the other before making all apps stop compiling

raulriera:
Stale bot 😬

alunbestor:
I'm all in favor of enums to make meaning explicit, though this would prevent us doing shit like:
```swift
view.property = state?.property ?? defaultValue
view.anotherProperty = (state != nil) ? customValue: defaultValue 
```
But further to what @TheoBdev said, the shorthands above are not intuitive to people who are new to the API.

"Use case" feels like the wrong term for this enum, but I didn't bring my paints for the bikeshed today.

I think it would be fairly straightforward to write a HostCell convenience init that wraps the old closures to work with the new API. Not sure if we can do the same at the ViewUpdater level.

PR #66, opened 2018-01-17
Cancel outstanding operations

krbarnes:
In some situations operations can pile up in the renderAndDiff queue. It's not necessary to run all of them because the last one added should provide the correct state to display. This PR cancels any outstanding operations that are not already executing before adding the new operation to the queue.

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -283,6 +283,10 @@ public class FunctionalTableData: NSObject {
 			
 			strongSelf.doRenderAndDiff(newSections, animated: animated, animations: animations, completion: completion)
 		}
+		renderAndDiffQueue.operations.forEach { (operation) in
+			guard operation.isExecuting == false else { return }
+			operation.cancel()
+		}
 		renderAndDiffQueue.addOperation(blockOperation)
 	}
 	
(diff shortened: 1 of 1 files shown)


FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -283,6 +283,10 @@ public class FunctionalTableData: NSObject {
 			
 			strongSelf.doRenderAndDiff(newSections, animated: animated, animations: animations, completion: completion)
 		}
+		renderAndDiffQueue.operations.forEach { (operation) in
```
g-Off: `renderAndDiffQueue.operations.lazy.filter { !$0.isExecuting }.forEach { $0.cancel() }`

krbarnes: Won’t this iterate over the array twice instead of once? Why lazy?

g-Off: lazy makes it iterate over elements only once :)
```
["A", "B", "CD", "EF"].lazy.filter { 
    print("filter") 
    return $0.count > 1 
}.forEach { 
    print("forEach") 
    $0.lowercased() 
}
```
spits out:
```
filter
filter
filter
forEach
filter
forEach
```

g-Off: (but I don't actually care, lazy is pretty cool though and makes for nice functional code without losing efficiency)

krbarnes: TIL.gif

etiennemartin:
What should we do about the callbacks passed in? There was a discussion about this before and I'm not sure it was resolved. Would there be any value in calling the completion block with some type of status enum (done, cancelled, failed?). The goal would be to provide more context to the callback to react to the completion accordingly?

Just a thought.

krbarnes:
The callbacks are still called on cancelled operations.

raulriera:
You should make a release 😜 (update the CHANGELOG, tag the framework)

PR #75, opened 2018-03-14
Expose supplementaryViewForElement in FunctionalCollectionData

josephroqueca:
There's currently no way to add headers and footers to a UICollectionView backed by FunctionalCollectionData, as FunctionalCollectionData controls the data source but does not expose `UICollectionViewDataSource.collectionView(_:viewForSupplementaryElementOfKind:at:)`.

This change exposes that method in the form of two functions -- `supplementaryElementReuseIdentifier` and `prepareViewForSupplementaryElement`. The first allows for a reuse identifier to be specified, which FunctionalCollectionData will use to dequeue a reusable cell, and then that cell will be passed through the second so attributes can be applied to it. This approach was taken because the data source *must* return a valid view, hence requiring the retrieval of the reuse identifier and dequeueing, and should not apply attributes (but still return it) if use of it is not desired, as specified [in the docs](https://developer.apple.com/documentation/uikit/uicollectionviewdatasource/1618037-collectionview), giving the option to avoid applying attributes through the second callback.

### DIFF
--- a/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
+++ b/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
@@ -70,6 +70,8 @@ public class FunctionalCollectionData: NSObject {
 	public var scrollViewDidEndScrollingAnimation: ((_ scrollView: UIScrollView) -> Void)?
 	public var scrollViewShouldScrollToTop: ((_ scrollView: UIScrollView) -> Bool)?
 	public var scrollViewDidScrollToTop: ((_ scrollView: UIScrollView) -> Void)?
+	public var supplementaryElementReuseIdentifier: ((_ collectionView: UICollectionView, _ kind: String, _ indexPath: IndexPath) -> String)?
+	public var prepareViewForSupplementaryElement: ((_ collectionView: UICollectionView, _ view: UICollectionReusableView, _ kind: String, _ indexPath: IndexPath) -> Void)?
 	
 	private let unitTesting: Bool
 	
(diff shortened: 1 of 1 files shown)


FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -456,6 +458,19 @@ extension FunctionalCollectionData: UICollectionViewDataSource {
...
+		// We expose this delegate method by requiring a method to get the reuseIdentifier, and then using that reuseIdentifier to
+		// dequeue a UICollectionReusableView, which can then have its appropriate attributes applied.
+		// This approach is required because `UICollectionViewDataSource` requires that this always return a valid view, and
+		// suggests skipping applying attributes to the view when a supplementary view is not wanted.
+		//
+		// The owner of the collectionView is still responsible for registering the supplementary element with `UICollectionView.register(_:forSupplementaryViewOfKind:withReuseIdentifier:)`
+		let reuseIdentifier = supplementaryElementReuseIdentifier?(collectionView, kind, indexPath) ?? ""
+		let view = collectionView.dequeueReusableSupplementaryView(ofKind: kind, withReuseIdentifier: reuseIdentifier, for: indexPath)
+		prepareViewForSupplementaryElement?(collectionView, view, kind, indexPath)
+		return view
```
raulriera: How do you intend to use this? I thought you wanted to have something like `TableHeaderFooterStateType` for UICollectionViews. This will only expose the guts of UICollectionDelegate to the view controller right? which goes "against" this library.

What are you going to be using the header for? (You can take this privately if its about implementation details)

PR #81, opened 2018-04-21
Add support for accessoryViews

jodyscho:
Updating the CellStyle to accept custom accessory views.

### DIFF
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -16,6 +16,7 @@ public struct CellStyle {
 	public var separatorColor: UIColor?
 	public var highlight: Bool?
 	public var accessoryType: UITableViewCellAccessoryType = .none
+	public var accessoryView: UIView?
 	public var selectionColor: UIColor? = CellStyle.selectionColor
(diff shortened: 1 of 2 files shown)


krbarnes:
I wonder if there’s a better way to do this. We added background view when iPhone X was released but I’ve been slowly removing uses of it with intention of deprecating. It’s really expensive to create and destroy views on every render pass.

jodyscho:
I'm open to different implementations but the ability to set the accessoryView simplifies client code. i.e. No need to create custom views to mimic an accessory.

As for being expensive, don't we have the same issue anyways. If a cell needs a custom view then one has to be created regardless.

Let me know what you guys decide.

jodyscho:
Oh, I see. You mean because we create all the HostCells in the render pass we'd potentially be creating a lot of views at that time.

jodyscho:
Maybe something like CombinedView that handles the accessory view. I'll give it some more thought.

raulriera:
I think these should be recreated using CombinedCells, this would be one more API in CellStyle that CollectionView wouldn’t be able to support.

PR #84, opened 2018-04-24
Host cell equality

Caiopia:
## What

This adds `key` and `style` to the equality definition of `CellConfigType`. While most style changes are accompanied by state changes, it's not always the case for indexed items that change positions. Adding style makes it more likely we update cells when they change their merged style. This also allows us to compare `CellConfigType`s more accurately outside of FunctionalTableData.

### Bonus ✨

- Allow demo app to be used with iOS 9.0+
- 🔥 1 warning

### DIFF
--- a/FunctionalTableData.xcodeproj/project.pbxproj
+++ b/FunctionalTableData.xcodeproj/project.pbxproj
@@ -497,7 +497,6 @@
 				DEVELOPMENT_TEAM = UD42KN75C7;
 				GCC_C_LANGUAGE_STANDARD = gnu11;
 				INFOPLIST_FILE = FunctionalTableDataDemo/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.3;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
 				PRODUCT_BUNDLE_IDENTIFIER = com.jadedpixel.FunctionalTableDataDemo;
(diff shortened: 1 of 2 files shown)


FunctionalTableData/HostCell.swift
```diff
@@ -95,7 +95,10 @@ public struct HostCell<View, State, Layout>: CellConfigType where View: UIView,
 	
 	public func isEqual(_ other: CellConfigType) -> Bool {
 		if let other = other as? HostCell<View, State, Layout> {
-			return state == other.state
+			var isEqual = key == other.key
+			isEqual = isEqual && state == other.state
+			isEqual = isEqual && style == other.style
+			return isEqual
```
krbarnes: I think it's ok to just `&&` the conditions together now.

krbarnes: Can you describe the issue that led you to here. The keys and styles are compared separately by TableSectionChangeSet so I would have thought that comparing state was enough here.

Caiopia: Main reason is that if we want to equate `CellConfigType`s outside of the `TableSectionChangeSet`, it wouldn't take `style` into account. Though I'm currently revisiting our use case to see if we really do need to equate these.

Caiopia:
Removing the HostCell equality, I don't need it anymore. Now it just removes a warning and the 11.3 demo app restriction.

PR #86, opened 2018-05-01
Example cleanup

raulriera:
Cleans up the example a bit and removed the CODEOWNERS

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,0 @@
-# Order is important; the last matching pattern takes the most precedence.
-* @raulriera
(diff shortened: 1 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -1,2 +0,0 @@
-# Order is important; the last matching pattern takes the most precedence.
-* @raulriera
```
Caiopia: Does this mean non-members can merge now too?

raulriera: It just means I won't be added as a reviewer all the time

krbarnes: _requests changes_

PR #90, opened 2018-05-15
Documented CellStyle

raulriera:
Added documentation for CellStyle. Now that we have support for `jazzy` docs, we should automate its generation every time we make changes https://github.com/Shopify/FunctionalTableData/commit/543d5108a15f6d9e56c79e6eeb43ba44fc7a04b6

The "meat" of this PR is on the first 2 commits, the 3rd one is generated documentation.


FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
 	public var selectionColor: UIColor? = CellStyle.selectionColor
+    /// The view's background color.
 	public var backgroundColor: UIColor?
+    /// The view that is displayed behind the cell’s other content.
 	public var backgroundView: UIView?
+    /// The tint color to apply to the cell.
 	public var tintColor: UIColor?
+    /// The default spacing to use when laying out content in the view.
 	public var layoutMargins: UIEdgeInsets?
+    /// The radius to use when drawing rounded corners for the layer’s background.
```
raulriera: what is this spacing? 😬

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
 import Foundation
 import UIKit
 
+/// The actions property exposed on the CellConfigType represents possible events that will be executed based on the users interaction with that particular cell. Of note are the `selectionAction` and `previewingViewControllerAction`. The `selectionAction` is executed when the user taps on that particular cell. The main use case for this is present a new detail view controller or a modal (but is not constrained to these actions, these are just the common use cases). The `previewingViewControllerAction` is responsible for returning an instance of a UIViewController that will be shown when a user 3D-touches on a cell.
```
alunbestor: This comment is a duplicate of the comment for `CellActions`, it does not document `CellStyle`.

Also line breaks are nice 🙈

Also missing apostrophe: `user's interaction`

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
 public struct CellStyle {
 	public static var selectionColor: UIColor? = nil // TODO: make this not a static like this
+	/// The style to apply to the bottom separator in the cell.
```
alunbestor: For all these separator properties I would say:
```
The style of separator line to display along the [top/bottom] of the cell. If `nil`, the containing `TableSection` may override the [top/bottom] separator based on its own `SeparatorStyle`.
```

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
+	///
+	/// Supported by UITableView only.
 	public var bottomSeparator: Separator.Style?
+	/// The style to apply to the top separator in the cell.
+	///
+	/// Supported by UITableView only.
 	public var topSeparator: Separator.Style?
+	/// The color of separator rows in the cell.
```
alunbestor: Clarification: `lines`, not `rows` (`rows` can be confused with the cells themselves.)

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
+	/// Whether the cell is highlighted or not.
+	///
+	/// Supported by UITableView only.
 	public var highlight: Bool?
+	/// The type of standard accessory control used by a cell.
+	/// You use these constants when setting the value of the [accessoryType](apple-reference-documentation://hspQPOCGHb) property.
+	///
+	/// Supported by UITableView only.
 	public var accessoryType: UITableViewCellAccessoryType = .none
+	/// The view's selection color.
```
alunbestor: Suggest:
```
The cell's background color while selected.
If `nil`, the cell's regular `backgroundColor` or `backgroundView` will be shown instead.
```
You could also move this property below `backgroundColor`.

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
 	public var selectionColor: UIColor? = CellStyle.selectionColor
+	/// The view's background color.
```
alunbestor: Suggest:
```
The background color to show behind the cell's content. Only used if `backgroundView` is `nil`.
```

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
 	public var backgroundColor: UIColor?
+	/// The view that is displayed behind the cell’s other content.
```
alunbestor: Suggest:
```
An optional view to display behind the cell's content. Overrides `backgroundColor` if specified.
```

FunctionalTableData/CellActions.swift
```diff
@@ -11,6 +11,7 @@ import UIKit
 
 /// The actions property exposed on the CellConfigType represents possible events that will be executed based on the users interaction with that particular cell. Of note are the `selectionAction` and `previewingViewControllerAction`. The `selectionAction` is executed when the user taps on that particular cell. The main use case for this is present a new detail view controller or a modal (but is not constrained to these actions, these are just the common use cases). The `previewingViewControllerAction` is responsible for returning an instance of a UIViewController that will be shown when a user 3D-touches on a cell.
 public struct CellActions {
+	/// The possible states a cell can be when a selection action is performed on it.
 	public enum SelectionState {
```
alunbestor: This comment does not reflect the purpose of the enum. Suggest:
```
Returned by `SelectionAction` to indicate whether the cell should be left selected or deselected after the user taps on it.
```

FunctionalTableData/CellStyle.swift
```diff
@@ -9,18 +9,41 @@
...
 	public var backgroundView: UIView?
+	/// The tint color to apply to the cell.
 	public var tintColor: UIColor?
+	/// The default spacing to use when laying out content in the view.
 	public var layoutMargins: UIEdgeInsets?
+	/// The radius to use when drawing rounded corners for the layer’s background.
```
alunbestor: Nice copypasta from the Apple docs :D but our `cornerRadius` doesn't just affect the background; `masksToBounds` causes the radius to also clip everything within the cell.

FunctionalTableData/CellStyle.swift
raulriera: I'll leave TableSection to handle that in their docs tho

FunctionalTableData/CellStyle.swift
raulriera: Did that for all the "iOS" properties

raulriera:
> 🤔Maybe mark all these docs files as autogen files in .gitattributes so that github will collapse them by default.

PRs welcome 😉

PR #93, opened 2018-05-17
More documentation

raulriera:
Fixed a typo in the documentation. Also added `Separator` into the mix


FunctionalTableData/Separator.swift
```diff
@@ -98,6 +119,9 @@ public extension UIView {
 		separator.constrainToBottomOfView(self)
 	}
 	
+	/// Removes any instance of a `Separator` view from the current view.
+	///
+	/// - Parameter withTag: the separator to remove.
```
dannbeaus: Just wondering since _ ; this parameter is not labeled for the consumer of this API - does it make sense to label it with the internal name in the docs?

raulriera: Yeah, they do (same example, official docs)

screen shot 2018-05-17 at 9 51 23 am

PR #95, opened 2018-05-20
tests aren't running on travis, using the script option kicks them off

g-Off:
Travis' YAML file configuration implies that tests run automatically, they don't. This kicks off the test phase using a script because even if you can get the config right it uses `xctool` by default.

How to validate:
If you look at previous test runs there is no build output and they all "succeed" in about 15s. This PR's CI output shows a build and test output and takes ~1.5mins

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: swift
+osx_image: xcode9.3
 xcode_project: FunctionalTableData.xcodeproj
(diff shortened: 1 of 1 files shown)


g-Off:
> Do we need to change | remove xcode_schema: FunctionalTableDataTests then?

It can probably be removed/doesn't do anything. I'm just going to leave it for now

PR #96, opened 2018-05-20
Added documentation to the init methods

raulriera:
Added documentation to the init methods, also removed those parameters. We may need to apply different default values

Also moved the example files into several groups for easier navigation


FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -81,9 +82,17 @@ public class FunctionalCollectionData: NSObject {
...
-		self.unitTesting = NSClassFromString("XCTestCase") != nil
-		self.name = name ?? "FunctionalCollectionDataRenderAndDiff-\((fileName as NSString).lastPathComponent):\(lineNumber)"
+	/// Initializes a FunctionalCollectionData. To configure its view, provide a UICollectionView after initialization.
+	///
+	/// - Parameter name: String identifying this instance of FunctionalCollectionData, useful when several instances are displayed on the same screen
+	public init(name: String? = nil) {
+		if let name = name {
+			self.name = name
+		} else {
+			self.name = "FunctionalCollectionDataRenderAndDiff-\((#file as NSString).lastPathComponent):\(#line)"
```
raulriera: Suggestions to change these #file and #line to something else? Obviously outside of the init they won't work. But as parameters they are also pretty strange

raulriera:
Bonus round adds documentation for UIScrollView methods

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
raulriera: Thoughts @g-Off ?

g-Off: UUID is even less useful. The file/line params indicate the creator of the instance of FunctionalTableData making it easier to trace back which instance may be generating a crash/exception

raulriera: In that case I think we should drop both then. If the name only, there is no value in both

g-Off: Might be able to change this to `name: String = "FunctionalCollectionDataRenderAndDiff-\(#file):\(#line)", which would give it a valid name all of the time and when one isn't provided would keep the file/line pairing

raulriera: Hmmmm apparently that doesn't work either, they have to be in their own argument to get the reference from the caller. Otherwise this

```swift
public init(name: String = "FunctionalTableDataRenderAndDiff-\((#file as NSString).lastPathComponent):\(#line)")
```

becomes, this

```shell
(lldb) po name
"FunctionalTableDataRenderAndDiff-FunctionalTableData.swift:144"
```

raulriera: No real reason apart from when I was documenting I went “hmmm what is this and why should someone care about these weird arguments”

I can put them back and **not** document them I gues. Best of both worlds?

PR #97, opened 2018-05-24
Apply default CellStyle when no custom style is provided

alunbestor:
Whenever a cell gets updated, `FunctionalXData` calculates the "merged" style for that cell (the combination of the cell's style and its section's style) and then asks that style to apply itself to the cell view using `CellStyle.configure(cell:)`

However:

`CellConfigType.style` and `TableSection.style` are optionals that default to `nil`, and when they're both `nil` then `TableSection.mergedStyle(for: cell)` also returns `nil`. **In this case, we don't apply *any* style to the cell and it retains its previous appearance.** This caused previous styles to be 'leaked': e.g. background colors, accessory types etc. could be left over from a previous style.

It also meant that certain default layout properties we assumed were always configured (e.g. `layoutMargins` and `preservesSuperviewLayoutMargins`) might never get applied for cells that don't have styles, causing cell behaviour to be unpredictable.

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -389,8 +389,8 @@ public class FunctionalTableData: NSObject {
 					update.cellConfig.update(cell: cell, in: tableView)
 					
 					let section = sections[update.index.section]
-					let style = section.mergedStyle(for: update.index.row)
-					style?.configure(cell: cell, in: tableView)
+					let style = section.mergedStyle(for: update.index.row) ?? CellStyle()
+					style.configure(cell: cell, in: tableView)
 				}
 			}
 		}
@@ -543,7 +543,8 @@ extension FunctionalTableData: UITableViewDataSource {
 		let cell = cellConfig.dequeueCell(from: tableView, at: indexPath)
 		
 		cellConfig.update(cell: cell, in: tableView)
-		sectionData.mergedStyle(for: row)?.configure(cell: cell, in: tableView)
+		let style = sectionData.mergedStyle(for: row) ?? CellStyle()
+		style.configure(cell: cell, in: tableView)
 		
 		return cell
 	}
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -389,7 +389,7 @@ public class FunctionalTableData: NSObject {
 					update.cellConfig.update(cell: cell, in: tableView)
 					
 					let section = sections[update.index.section]
-					let style = section.mergedStyle(for: update.index.row) ?? CellStyle()
+					let style = section.mergedStyle(for: update.index.row)
 					style.configure(cell: cell, in: tableView)
 				}
 			}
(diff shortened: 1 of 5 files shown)


alunbestor:
This edge case was probably what caused the elusive spacer-color bug in Mobile Shopify - a spacer with a custom background color applied by a CellStyle was getting reused for a spacer with no cell style that wanted the default background color.

I'm not sure of this though, and I believe that bug was already fixed by an unrelated change that made plain and gradient spacers use entirely different types (which would not reuse the same views.)

alunbestor:
I don't know the context behind making `CellConfigType.style` optional, rather than making it required and defaulting to `CellStyle`, the way we handle `CellActions`. It already kinda-sorta-accidentally defaults to `CellStyle` anyway: if the section has a style but the cell does not, then `mergedStyle` will build the resulting style from `CellStyle()`.

Can someone speak to the decision to make it nil? Would it be better to make the change at the `CellConfigType` level now and enforce non-nullability throughout?

raulriera:
Why not fix this in HostCell instead? Can the issue be reproches in tests?

alunbestor:
> Why not fix this in HostCell instead?

This is what I meant about `CellConfigType.style` being marked as optional - if we made it non-nil then we could fix this at the `HostCell` level. That's my preference, though that's a source-breaking change, and I want more context on why it was `nil` to start with.

Right now, AFAICT the optionality is just making the behaviour less defined rather than adding more flexibility.

> Can the issue be reproduced in tests?

Good point, I'll work on some tests to demonstrate the bug.

raulriera:
HostCell is just one of many possible implementations of CellConfigType. Just set it to = CellStyle in the init?

alunbestor:
> HostCell is just one of many possible implementations of CellConfigType. Just set it to = CellStyle() in the init?

Sure, but FunctionalTableData/FunctionalCollectionDats are responsible for applying the style, not HostCell; if `CellConfigType.style` remains nil they would still have to account for that nullability (and the ambiguous behaviour it introduces), even if we made the concrete `HostCell` implementation always return a non-nil `CellStyle` for its (still-optional) `style` property.

alunbestor:
What were our feelings about this?

raulriera:
Travis-CI has feelings

alunbestor:
IDK how to get Travis-CI to unshit the bed, I couldn't find a Retry button 😔

alunbestor:
Didn't mean for this to go stale - I'm reopening it and will try to write some unit tests for it. This bug has come up again in https://github.com/Shopify/ios/pull/5076, where leftover disclosure indicators were getting reused on cells that shouldn't have them.

I'd like some eyes on the discussion around making `CellConfigType.style` non-optional. This would be a source-breaking change for downstream projects.

FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -386,8 +386,8 @@ public class FunctionalTableData: NSObject {
 					update.cellConfig.update(cell: cell, in: tableView)
 					
 					let section = sections[update.index.section]
-					let style = section.mergedStyle(for: update.index.row)
-					style?.configure(cell: cell, in: tableView)
+					let style = section.mergedStyle(for: update.index.row) ?? CellStyle()
```
g-Off: I'd make `mergedStyle` return a non-optional and have it return the `CellStyle()`

alunbestor:
I've added a unit test to test that previous styles are not left over when a cell is reused, and made `mergedStyle` return a non-optional result.

I played with making `CellConfigType.style` non-optional as well, but it required a lot of downstream code changes. I still think that's the best approach, but we can look at that in a separate PR.

alunbestor:
Bump. This PR is totally not dead, it's ready for a re-review.

alunbestor:
Could I get (do I need) another review for this?

PR #98, opened 2018-05-25
Support for edges of a cell

raulriera:
This changes the way we specify the layout of a `CellConfigType`. Previously we were forced to use a pair of the edges together, or all four of them. Now we have control over each.

Also removes a bunch of auto layout code we don't need anymore


FunctionalTableData/TableItemLayout.swift
```diff
@@ -14,44 +14,117 @@ public protocol TableItemLayout {
...
+	
+	public struct Leading: TableItemLayout {
+		public static func layoutView(_ view: UIView, inContentView contentView: UIView) {
+			view.leadingAnchor.constraint(equalTo: contentView.leadingAnchor).isActive = true
+		}
+	}
+	
+	public struct Bottom: TableItemLayout {
+		public static func layoutView(_ view: UIView, inContentView contentView: UIView) {
+			contentView.bottomAnchor.constraint(equalTo: view.bottomAnchor).isActive = true
```
alunbestor: It's awkward that these definitions flip the order of `view` and `contentView` for the bottom and trailing constraints. All four should be consistently defined.

FunctionalTableData/UIView+AutoLayout.swift
```diff
@@ -15,88 +15,4 @@ extension UIView {
 			addSubview(view)
 		}
 	}
-	
-	func constraintsToFillView(_ otherView: UIView, respectingLayoutMargins: Bool = false) -> [NSLayoutConstraint] {
```
alunbestor: 🔥I'm glad all these were never `public`.

FunctionalTableData/TableItemLayout.swift
raulriera: > It's awkward that these definitions flip the order of view and contentView for the bottom and trailing constraints. All four should be consistently defined.

This is how Interface Builder does it as well (flipping the order), otherwise we will have to `* -1` any value passed to it.

In this cases is safe because there is no constant added

alunbestor: Had not noticed that IB was doing that too 🤔 I guess for the same reason you describe.

raulriera:
> However depending on how much time we want to invest to make it fully compatible we could consider not removing constrainToFillViewHorizontally & such but just deprecating it.

Thankfully all of that are just internal helper methods that we don't need anymore 🙂 I was eager to delete them, finally was able to 😄

PR #100, opened 2018-06-14
Add hooks for reporting accessibility scroll status

alunbestor:
This PR adds an optional callback for `UIAccessibilityScrollViewDelegate.accessibilityScrollStatus`, which can be used to report the current scroll position of the table in an accessible way: e.g. "Page 2 of 3" (for general screen content), "Aardvark to Marmoset" (for alphabetised lists), "15th January to 23rd March" (for chronological lists) etc. This is a useful navigational aid for VoiceOver users.

This PR also adds a helper method to get all currently-visible cells, which proved useful for constructing these accessibility descriptions.

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -105,6 +105,10 @@ public class FunctionalTableData: NSObject {
 	public var scrollViewShouldScrollToTop: ((_ scrollView: UIScrollView) -> Bool)?
 	/// See UIScrollView's [documentation](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619382-scrollviewdidscrolltotop) for more information.
 	public var scrollViewDidScrollToTop: ((_ scrollView: UIScrollView) -> Void)?
+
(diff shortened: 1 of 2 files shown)


FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -153,6 +157,15 @@ public class FunctionalTableData: NSObject {
...
 	}
 	
+	/// All rows that are currently visible in the table.
+	public var visibleRows: [CellConfigType] {
+		guard let visibleIndexPaths = tableView?.indexPathsForVisibleRows else {
+			return []
+		}
+		
+		return visibleIndexPaths.compactMap { sections[$0] }
+	}
```
raulriera: What value are you getting out of the CellConfigType? are you then casting those to their actual types? I think this bit property be outside of the library tho

alunbestor:
> This is cool, do we have plans to incorporate it for accessibility work internally?

Yup, once this is merged to FTD I have a Polaris PR ready to go up that'll implement accessible scroll descriptions.

alunbestor:
I've rebased, removed the `visibleRows` helper, and added the equivalent callback to `FunctionalCollectionData`. Will merge on 🍏.

PR #101, opened 2018-06-14
Supply duplicate keys in exception user info

sherryshao:
In an attempt to send FTD crash attachments for POS (https://github.com/Shopify/ios/pull/4133), I've come across some exception reasons we could improve to make a crash log clearer.

Proposing to supply only duplicate keys in both `userInfo` as well as attaching it to the `reason` text of an exception so that a crash log can be deciphered faster without having to dig into the `*.tabledata` attachment (that's still very useful, which is why we're also making changes on POS to attach that upon crashes, but maybe for more obscure bugs).


raulriera:
Looks like indentation is all over the place. Apart from that, what is the reasoning for this? Only faster debugging purposes? This is something that should be applied to the `.tabledata` itself (as a parser tool, independent of the Swift code) and not where you are doing it in my opinion, because this is making renderAndDiff slower, since its doing all the calculations all the time, even when nothing is duplicated

sherryshao:
@raulriera so we do already do a `.map` previously, every time we render and diff, whereas now we do a `.forEach`, which is both O(n) in terms of complexity (I guess now it's more like O(2n), but we can still get it down to O(n) if `TableSection` and `CellConfigType` both just implement their duplicate key implementations  - I don't actually think we're sacrificing performance with this implementation - just a different output for the spending the same time complexity.

raulriera:
> so we do already do a .map previously

True, but now we do a `.map` a `.forEach` a `.contains` for everything, on large lists (order lists, this would be totally unnecessary) 

> it's true that *.tabledata will contain all the info we need, but depending on how big your table is, it could still be a bit of a hassle to dig through and find the right section or row keys that happen to duplicate. So, if we can spend the same time complexity and have a more verbose reason output, I'd vote for that granted the FTD team agrees as well.

What I meant is create an script, app, or anything that reads the `.tabledata` and massages it to the POS needs for debugging?

Other problems is that this exposes "debug" tings like `duplicatedKeys` into TableSection. Where it shouldn't even be possible (because results in a crash), so that property shouldn't exist.

sherryshao:
Since the extension is for `Hashable`s and we’re calling `.contains` on a set, O(1) would be the complexity. I can definitely remover the extraneous `.map` and do a for each on the individual section or row objects themselves. That way we really just have O(n).

 I guess I’m just speaking from a user perspective from debugging with FTD and how I feel like there could be improvements that I could just reflect directly on our crash report - I don’t think always diff-ing the `.tabledata` as part of our debugging process is the most helpful alternative every time we get an FTD crash (given the same time complexity). 

I agree that maybe this logic doesn’t belong in the main FTD file, I could definitely find other places where this could belong - what do you think?

sherryshao:
@raulriera just going to supply in user info and pick it off the exception handler from our end as @g-Off had mentioned the crash groupings would be a bit messed up if we choose to do this. I opted for only generating duplicate keys as we're about to raise an exception, and not supply it in the reason at all.

sherryshao:
Looks like indentation oddness was caused by me using spaces instead of tabs locally 🙈

sherryshao:
@raulriera @g-Off would appreciate another 👀 when you have time 🙏

FunctionalTableData/Array+TableSection.swift
```diff
@@ -0,0 +1,56 @@
...
+				let reason = "\(senderName) : Duplicate Section.Row keys"
+				let userInfo: [String: Any] = ["Section": section.key, "Duplicates": dupKeys]
+				NSException(name: NSExceptionName.internalInconsistencyException, reason: reason, userInfo: userInfo).raise()
+			}
+		}
+	}
+}
+
+private extension Array where Element: Hashable {
+	func duplicates() -> Set<Element> {
```
raulriera: We could use https://developer.apple.com/documentation/swift/set/2853807-intersection instead

FunctionalTableData/Array+TableSection.swift
```diff
@@ -0,0 +1,56 @@
+//
+//  Array+TableSection.swift
+//  FunctionalTableData
+//
+//  Created by Sherry Shao on 2018-06-14.
+//  Copyright © 2018 Shopify. All rights reserved.
+//
+
+extension Array where Element: TableSectionType {
```
raulriera: Can you put this file in the extensions folder?

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -36,9 +36,9 @@ public class FunctionalCollectionData: NSObject {
 		}
 	}
 	
-	private func dumpDebugInfoForChanges(_ changes: TableSectionChangeSet, previousSections: [TableSection], visibleIndexPaths: [IndexPath], exceptionReason: String?) {
+	private func dumpDebugInfoForChanges(_ changes: TableSectionChangeSet, previousSections: [TableSection], visibleIndexPaths: [IndexPath], exceptionReason: String?, exceptionUserInfo: [AnyHashable: Any]?) {
 		guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
-		let exception = FunctionalTableData.Exception(name: name, newSections: sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: collectionView?.frame ?? .zero, reason: exceptionReason)
```
raulriera: Why AnyHashale?

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -507,7 +489,7 @@ extension FunctionalCollectionData: UICollectionViewDelegate {
 			}
 		}
 	}
-
+
```
raulriera: Can you fix all of these please?

FunctionalTableData/FunctionalTableData.h
```diff
@@ -27,4 +27,13 @@ static inline void catchAndRethrowException(__attribute__((noescape)) void (^ __
 	}
 }
 
+__attribute__((visibility("hidden")))
+static inline void catchException(__attribute__((noescape)) void (^ __nonnull inBlock)(void), __attribute__((noescape)) void (^ __nonnull exceptionHandler)(NSException *)) {
```
raulriera: Do we need this?

FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -113,7 +114,7 @@ public class FunctionalTableData: NSObject {
 			let reload: UITableViewRowAnimation
 			public static let `default` = Actions(insert: .fade, delete: .fade, reload: .automatic)
 			public static let legacy = Actions(insert: .top, delete: .top, reload: .automatic)
-
+
```
raulriera: Same about these

FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -660,19 +643,19 @@ extension FunctionalTableData: UITableViewDelegate {
 	public func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
 		guard let cell = tableView.cellForRow(at: indexPath) else { return }
 		let cellConfig = sections[indexPath]
-
+		
 		let selectionState = cellConfig?.actions.selectionAction?(cell) ?? .deselected
```
raulriera: Same

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
sherryshao: it's the default `userInfo` type I believe that you get from exceptions

FunctionalTableData/FunctionalTableData.h
sherryshao: Just for testing purposes, I could move that over to the test suite too if you'd like - instead of rethrowing, we'd just handle it (ex. fail or assert otherwise). Thought it was generic enough to stay in the main codebase.

FunctionalTableData/Array+TableSection.swift
sherryshao: unfortunately the actual content we care about is an array (i.e. the subject of this extension), and the duplicates in said array. 

`intersection` could only be called from a set, and while we can supply a sequence to it, calling intersect on the unique keys and supplying this array will only yield the same unique keys set.

We could also do:
```
var uniqueKeys = Set(allKeys)
return allKeys.reduce(into: Set<Int>()) {
    if uniqueKeys.contains($1) {
        uniqueKeys.remove($1)
    } else {
        $0.insert($1)
    }
}
```

I just found the current version easier to read.

raulriera: I meant internally, you are doing the exact thing intersection is doing, finding the duplicates in the sets

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
raulriera: True, but do we need it tho? You are only creating dictionaries with keys as string `String:Any`

sherryshao: Sure - was just being general, I can change that.

FunctionalTableData/Array+TableSection.swift
sherryshao: So and intersection will find the common elements between two sets. What we have here is an array with duplicate elements in and of itself, if we turn it into a set, we lose that information.

Maybe I’m not understanding you correctly, do you mind giving an example?

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
sherryshao: Actually, these appeared because I reformatted the file with tabs locally - if you take a look at the rest of the file, 
all extra lines will have this indentation - this is actually the correct format

FunctionalTableData/Array+TableSection.swift
sherryshao: Fun fact, there never was a physical extension folder, just a project grouping called Extensions. I can create one.

FunctionalTableData/CollectionView/FunctionalCollectionData.swift
sherryshao: Actually, since the way we're passing info back and forth in FTD and FCD is via a native exception, which passes its own `userInfo`, I think it's better that we just adhere to its type without having to convert back and forth.

sherryshao:
@raulriera comments addressed, otherwise I've added an extension folder to the ftd project.

FunctionalTableData/TableView/FunctionalTableData.swift
sherryshao: See comment above regarding indentation in the rest of the file

FunctionalTableData/TableView/FunctionalTableData.swift
sherryshao: ^

FunctionalTableData/Array+TableSection.swift
raulriera: Oh never mind sorry, I thought you were looking for the duplicates between `sections` and `newSections`

PR #102, opened 2018-06-19
Make FTD and FCD conform to UIScrollViewAccessibilityDelegate

alunbestor:
Oops. In https://github.com/Shopify/FunctionalTableData/pull/100 I forgot to make `FunctionalTableData` and `FunctionalCollectionData` actually advertise that they conform to `UIScrollViewAccessibilityDelegate`, which meant `accessibilityScrollStatus` was never called. This PR fixes that.

### DIFF
--- a/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
+++ b/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
@@ -586,9 +586,11 @@ extension FunctionalCollectionData: UICollectionViewDelegate {
 	public func scrollViewDidScrollToTop(_ scrollView: UIScrollView) {
 		scrollViewDidScrollToTop?(scrollView)
 	}
-	
-	// MARK: - UIScrollViewAccessibilityDelegate
-	
+}
+
+// MARK: - UIScrollViewAccessibilityDelegate
+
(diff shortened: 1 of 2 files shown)


FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -586,9 +586,11 @@ extension FunctionalCollectionData: UICollectionViewDelegate {
...
 		scrollViewDidScrollToTop?(scrollView)
 	}
-	
-	// MARK: - UIScrollViewAccessibilityDelegate
-	
+}
+
+// MARK: - UIScrollViewAccessibilityDelegate
+
+extension FunctionalCollectionData: UIScrollViewAccessibilityDelegate {
```
RyanBillard: how do these actually get called? Does UIScrollView internally check to see if its delegate conforms to this?

alunbestor: Apparently so. `UIScrollView` tests if its delegate conforms to the protocol, and if it doesn't then it won't bother checking if the method itself has been implemented. Hence why the previous implementation (which was grouped in with the `UIScrollViewDelegaate` methods) wasn't being called.

PR #106, opened 2018-08-14
Remove my personal team id

krbarnes:
Never realized this caused signing errors because I have the correct certificate, but I'm the only person that does.


dannbeaus:
Apparently if you dont 🔥 derived data it doesn't show as an issue.

PR #107, opened 2018-08-14
Change changeset to class

krbarnes:
Changing `TableSectionChangeSet` from a struct to class.

Since Swift 4.1 we have seen many crashes in `HostCell.isEqual`. None of the crashes make sense, but the crash is a SEGV_ACCERR which implies that multiple threads are accessing the same memory. 

Changing the change set from struct to class and eliminating the mutating functions *_maybe_* makes these crashes go away.

### DIFF
--- a/FunctionalTableData/HostCell.swift
+++ b/FunctionalTableData/HostCell.swift
@@ -93,7 +93,6 @@ public struct HostCell<View, State, Layout>: CellConfigType where View: UIView, 
 		cellUpdater(cell.view, state)
 	}
 	
-	@inline(never)
 	public func isEqual(_ other: CellConfigType) -> Bool {
 		if let other = other as? HostCell<View, State, Layout> {
 			return state == other.state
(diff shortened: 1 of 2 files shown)


FunctionalTableData/HostCell.swift
```diff
@@ -93,7 +93,6 @@ public struct HostCell<View, State, Layout>: CellConfigType where View: UIView,
 		cellUpdater(cell.view, state)
 	}
 	
-	@inline(never)
```
krbarnes: This was a previous attempt to fix the same crash. It did nothing.

FunctionalTableData/TableSectionChangeSet.swift
```diff
@@ -19,7 +19,7 @@ struct Moved<T: Equatable>: Equatable {
 
 // Compares two arrays of TableSections and produces the operations
 // required to go from one to the other
-public struct TableSectionChangeSet {
+public class TableSectionChangeSet {
```
raulriera: `final`? ❤️

krbarnes: sure

dannbeaus: finally

PR #108, opened 2018-08-21
Fix failing typecast in section validation tests

alunbestor:
This PR fixes a conditional cast from `[String]` to `Set<String>` that now (as of Swift 4.2 maybe?) always fails, which was causing a pair of tests to fail.

### DIFF
--- a/FunctionalTableDataTests/TableSectionsValidationTests.swift
+++ b/FunctionalTableDataTests/TableSectionsValidationTests.swift
@@ -37,7 +37,12 @@ class TableSectionsValidationTests: XCTestCase {
 		}, failure: {
 			XCTAssertEqual($0.name, NSExceptionName.internalInconsistencyException)
 			XCTAssertNotNil($0.userInfo as? [String: Any])
-			XCTAssertEqual($0.userInfo!["Duplicates"] as? Set<String>, Set(arrayLiteral: "section1"))
+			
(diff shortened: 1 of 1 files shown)


FunctionalTableDataTests/TableSectionsValidationTests.swift
```diff
@@ -37,7 +37,12 @@ class TableSectionsValidationTests: XCTestCase {
 		}, failure: {
 			XCTAssertEqual($0.name, NSExceptionName.internalInconsistencyException)
 			XCTAssertNotNil($0.userInfo as? [String: Any])
-			XCTAssertEqual($0.userInfo!["Duplicates"] as? Set<String>, Set(arrayLiteral: "section1"))
+			
+			if let duplicates = $0.userInfo?["Duplicates"] as? [String] {
+				XCTAssertEqual(Set(duplicates), Set(["section1"]))
```
raulriera: isn't this always going to make them the same? if duplicates is `["section1", "section1"]` it will be collapsed to `["section1"`]

alunbestor: `duplicates` is expected to contain each unique key that was duplicated: one entry per key, not one per section, but we don't care about the number. What we really care about is "does duplicates contain the duplicate key, with no false positives for other keys?"

PR #109, opened 2018-09-14
Set cell accessibility identifiers to section+cell keys

Caiopia:
## What

Adds an accessibility identifier to each cell so we can access them more easily for UI Testing. Since we already add unique keys to each section configuration and cell configuration, we should be able to reuse them as the accessibility identifiers. This way we don't have to input any extra identifiers.

If you have two components inside a cell you want to access separately you can set those components as accessibility elements to override the cell's accessibility identifier.

Since accessibility identifiers are only used for UITesting, this shouldn't affect any integrations with Voice Over.

### Example

Similar to the test, if we had a section:
```swift
TableSection(key: "sectionKey", rows: [SomeCell(key: "cellKey")])
```
The `accessibilityIdentifier` of the cell inside that section will be `"sectionKeycellKey"`

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -541,6 +541,7 @@ extension FunctionalTableData: UITableViewDataSource {
 		let row = indexPath.row
 		let cellConfig = sectionData[row]
 		let cell = cellConfig.dequeueCell(from: tableView, at: indexPath)
+		cell.accessibilityIdentifier = sectionData.sectionKeyPathForRow(row)
 		
 		cellConfig.update(cell: cell, in: tableView)
 		sectionData.mergedStyle(for: row)?.configure(cell: cell, in: tableView)
--- a/FunctionalTableDataTests/FunctionalDataTests.swift
+++ b/FunctionalTableDataTests/FunctionalDataTests.swift
@@ -73,6 +73,25 @@ class FunctionalDataTests: XCTestCase {
 		XCTAssertNotNil(indexPath)
 	}
 	
+	func testCellAccessibilityIdentifiers() {
+		let tableData = FunctionalTableData()
+		let tableView = UITableView()
+		
+		tableData.tableView = tableView
+		let expectation1 = expectation(description: "rendered")
+		let cellConfig = TestCaseCell(key: "cellKey", state: TestCaseState(data: "red"), cellUpdater: TestCaseState.updateView)
+		let section = TableSection(key: "sectionKey", rows: [cellConfig])
+		
+		tableData.renderAndDiff([section]) {
+			expectation1.fulfill()
+		}
+		waitForExpectations(timeout: 1, handler: nil)
+		
+		let cell = tableData.tableView?.visibleCells.first
+		XCTAssertNotNil(cell)
+		XCTAssertEqual(cell!.accessibilityIdentifier, section.sectionKeyPathForRow(0))
+	}
+	
 	func testPerformance() {
 		let functionalData = FunctionalTableData()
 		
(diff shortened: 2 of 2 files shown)


FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -541,6 +541,7 @@ extension FunctionalTableData: UITableViewDataSource {
 		let row = indexPath.row
 		let cellConfig = sectionData[row]
 		let cell = cellConfig.dequeueCell(from: tableView, at: indexPath)
+		cell.accessibilityIdentifier = sectionData.sectionKeyPathForRow(row)
```
raulriera: Not horrible feelings, but I believe this is the job of the states. Cells can be very complex with a lot of views, and we will need to identify each of them, like we are currently doing in some of our states `ButtonState` for example.

ArielSD: Meaning that the HostCell's state should hold the identifier? And for the views inside, couldn't we just set their identifiers to override the cell as the superview like Caio said?

raulriera: Yes, that is what I mean. I have no strong opinions on it, but I dont think the key makes a great identifier. Not every view needs an unique identifier, I would argue all orders cell for example should have the same identifier. But I dont know the scope of this change

Caiopia: That information should still belong in `ButtonState` along with setting `accessibilityLabel`s since we might want to access that button specifically in a UI test even if it's contained inside a HostCell's view, but having cell identifiers in addition would be super helpful. With this, we can add support for cell selection in UI tests for free and differentiate from other accessible items that are repeated in similar cells more easily.

> I have no strong opinions on it, but I dont think the key makes a great identifier. Not every view needs an unique identifier, I would argue all orders cell for example should have the same identifier.

Having unique identifiers for things like individual orders has been pretty useful so far. If you want to access the 4th order, you can use the key you already wrote instead of getting the cell at index 3. I can write up an example...

> But I dont know the scope of this change

`accessibilityIdentifier` is only used for UI testing, so this is an additive change for those doing UITesting. If you want your state to set your hosted view's accessibility information, you can still do that the same way as previously, but the cells that contain it will have identifiers instead of being anonymous.

alunbestor: As others have said this is purely additive - you can still access elements in a set by index, that's not changing.

For list views, maybe we shouldn't write UI tests that target rows by ID. But for detail views, the keypath usually carries semantic meaning and is a useful way of identifying particular pieces of the UI.

This change gives us the flexibility to use both approaches.

FunctionalTableDataTests/FunctionalDataTests.swift
```diff
@@ -73,6 +73,25 @@ class FunctionalDataTests: XCTestCase {
...
+		let section = TableSection(key: "sectionKey", rows: [cellConfig])
+		
+		tableData.renderAndDiff([section]) {
+			expectation1.fulfill()
+		}
+		waitForExpectations(timeout: 1, handler: nil)
+		
+		let cell = tableData.tableView?.visibleCells.first
+		XCTAssertNotNil(cell)
+		XCTAssertEqual(cell!.accessibilityIdentifier, section.sectionKeyPathForRow(0))
```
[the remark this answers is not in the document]
alunbestor: `cell?.accessibilityIndentifier`. XCTAssertEqual handles optionals just fine, and this prevents the test from crashing if `XCTAssertNotNil(cell)` fails.

Agree with Raul though that it should use a hardcoded comparison value. That documents the format of the expected output, on which all of our unit tests will depend, so if it changes we want to know about it.

Caiopia:
Updated test to use hard coded accessibility identifier.

PR #111, opened 2018-09-19
Remove hardcoded white background color from CellStyle

alunbestor:
Previously, `CellStyle` would force a cell's background color to white if no specific background color was provided. This is an assumption that a public framework shouldn't be making, and it prevents you from defining cells with no background color at all.

This PR:
- moves the default color to a static `CellStyle` property that can be overridden by client apps (including setting it to `nil`.)
- allows cells to have no background color.
- avoids creating a background view on every cell update just to hold the color.
- makes the background color (if any) stack with the background view (if any).

### DIFF
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -13,7 +13,22 @@ import UIKit
 ///
 /// Some properties are only supported by `UITableView`.
 public struct CellStyle {
-	public static var selectionColor: UIColor? = nil // TODO: make this not a static like this
+	/// The default selection color to use for styles that don't specify their own.
+	public static var defaultSelectionColor: UIColor? = nil
+	
+	/// The default background color to use for styles that don't specify their own.
+	public static var defaultBackgroundColor: UIColor? = .white
+	
+	@available(*, deprecated, message: "Renamed to defaultSelectionColor.")
+	public static var selectionColor: UIColor? {
+		get {
+			return defaultSelectionColor
+		}
+		set {
+			defaultSelectionColor = newValue
+		}
+	}
+	
 	/// The style to apply to the bottom separator in the cell.
 	///
 	/// Supported by `UITableView` only.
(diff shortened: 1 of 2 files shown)


FunctionalTableData/CellStyle.swift
```diff
@@ -34,9 +49,9 @@ public struct CellStyle {
 	/// You use these constants when setting the value of the [accessoryType](apple-reference-documentation://hspQPOCGHb) property.
 	///
 	/// Supported by `UITableView` only.
-	public var accessoryType: UITableViewCell.AccessoryType = .none
+	public var accessoryType: UITableViewCell.AccessoryType
 	/// The view's selection color.
-	public var selectionColor: UIColor? = CellStyle.selectionColor
```
alunbestor: These property-level defaults were never used, since the public `init` touches every property and overrides them with its parameter defaults.

FunctionalTableData/CellStyle.swift
```diff
@@ -13,7 +13,22 @@ import UIKit
 ///
 /// Some properties are only supported by `UITableView`.
 public struct CellStyle {
-	public static var selectionColor: UIColor? = nil // TODO: make this not a static like this
+	/// The default selection color to use for styles that don't specify their own.
+	public static var defaultSelectionColor: UIColor? = nil
```
alunbestor: I know we said "hey let's not do this", but I'm not averse to this approach; it's a pragmatic way of getting UIAppearance-style behaviour on a struct.

FunctionalTableData/CellStyle.swift
```diff
@@ -73,14 +88,8 @@ public struct CellStyle {
...
-		if let backgroundView = backgroundView {
-			cell.backgroundView = backgroundView
-		} else {
-			cell.backgroundColor = backgroundColor ?? UIColor.white
-			let backgroundView = UIView()
-			backgroundView.backgroundColor = cell.backgroundColor
-			cell.backgroundView = backgroundView
-		}
+		cell.backgroundColor = backgroundColor
+		cell.backgroundView = backgroundView
```
krbarnes: UITableViewCell.backgroundView is nil in vanilla UIKit by default right? We're not nilling out something that had a value now?

alunbestor: Did some testing with a tableview created in code in a new project:

- `UITableViewCell.backgroundView` is nil by default and remains nil when the cell appears.
- `UITableViewCell.backgroundColor` is "nil" by default, but gets set to `UIColor.white` at some point between `tableView(_:cellForRowAt:)` and when the cell appears.
- ...*except* if you set the color manually to `nil` in `tableView(_:cellForRowAt:)`, then it remains `nil` when the cell appears.

So giving `CellStyle` a default background color of white (but letting you override it to nil) mimics how UIKit works out of the box.

FunctionalTableData/CellStyle.swift
dannbeaus: oh I read that, and completely blanked after that.
🙆‍♂️

PR #112, opened 2018-10-04
Add BackgroundViewProvider and deprecate CellStyle.backgroundView.

Campbell-Scott:
The `backgroundView` property on `CellStyle` is too expensive - background views are created on every render pass, even though they are rarely used. 

The `BackgroundViewProvider` allows `CellStyle` to request an optional background view when its being dequeued; the only time it's actually needed.

### DIFF
--- a/FunctionalTableData/BackgroundViewProvider.swift
+++ b/FunctionalTableData/BackgroundViewProvider.swift
@@ -19,16 +19,14 @@ public protocol BackgroundViewProvider {
 	func isEqualTo(_ other: BackgroundViewProvider) -> Bool
 }
 
-public extension BackgroundViewProvider {
-	public func isEqualTo(_ other: BackgroundViewProvider) -> Bool {
-		return type(of: self) == type(of: other)
-	}
-}
-
 public struct EmptyBackgroundProvider: BackgroundViewProvider {
 	public init() {}
 
 	public func backgroundView() -> UIView? {
 		return nil
+	}
+
+	public func isEqualTo(_ other: BackgroundViewProvider) -> Bool {
+		return type(of: self) == type(of: other)
 	}
 }
--- a/FunctionalTableData/BackgroundViewProvider.swift
+++ b/FunctionalTableData/BackgroundViewProvider.swift
@@ -1,0 +1,34 @@
+//
+//  BackgroundViewProvider.swift
+//  FunctionalTableData
+//
+//  Created by Scott Campbell on 2018-10-16.
+//  Copyright © 2018 Shopify. All rights reserved.
+//
+
+import UIKit
+
+/// Provide a background view to be displayed behind the other contents of a cell.
+/// An implementation should maintain some internal state about the contents of the view.
+public protocol BackgroundViewProvider {
+	/// This is where the background view should be instantiated, since this
+	/// function is only called when a cell is being prepared to be deqeueued.
+	func backgroundView() -> UIView?
+
+	/// Compare the internal state to avoid unnecessarily instantiating the background view.
+	func isEqualTo(_ other: BackgroundViewProvider) -> Bool
+}
+
+public extension BackgroundViewProvider {
+	public func isEqualTo(_ other: BackgroundViewProvider) -> Bool {
+		return type(of: self) == type(of: other)
+	}
+}
+
+public struct EmptyBackgroundProvider: BackgroundViewProvider {
+	public init() {}
+
+	public func backgroundView() -> UIView? {
+		return nil
+	}
+}
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -54,8 +54,11 @@ public struct CellStyle {
 	public var selectionColor: UIColor?
 	/// The view's background color.
 	public var backgroundColor: UIColor?
-	/// The view that is displayed behind the cell’s other content.
+	/// The view that is displayed behind the cell's other content.
+	@available(*, deprecated, message: "Replaced with backgroundViewProvider.")
 	public var backgroundView: UIView?
+	/// Provides the view that is displayed behind the cell's other content.
+	public var backgroundViewProvider: BackgroundViewProvider
 	/// The tint color to apply to the cell.
 	public var tintColor: UIColor?
 	/// The default spacing to use when laying out content in the view.
@@ -63,6 +66,7 @@ public struct CellStyle {
 	/// The radius to use when drawing rounded corners in the view.
 	public var cornerRadius: CGFloat
 	
+	@available(*, deprecated, message: "The `backgroundView` argument is no longer available. Use backgroundViewProvider instead.")
 	public init(topSeparator: Separator.Style? = nil,
 	            bottomSeparator: Separator.Style? = nil,
 	            separatorColor: UIColor? = nil,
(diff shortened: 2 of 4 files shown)


FunctionalTableData/CellStyle.swift
```diff
@@ -66,14 +67,38 @@ public struct CellStyle {
...
 		self.backgroundColor = backgroundColor
-		self.backgroundView = backgroundView
+		self.backgroundViewProvider = BackgroundViewProvider({ return backgroundView })
+		self.tintColor = tintColor
+		self.layoutMargins = layoutMargins
+		self.cornerRadius = cornerRadius
+	}
+
+	public init(topSeparator: Separator.Style? = nil,
+				bottomSeparator: Separator.Style? = nil,
```
dannbeaus: weird indents compared to above.

Campbell-Scott:
@dannbeaus every time we tell a builder to create all the sections and their cells for a diff, if a cell has a backgroundView, we create and destroy all those views, even for cells that aren't visible. With a provider, we can provide a backgroundView on-demand when the view is actually going to be rendered. This way we stop ourselves from creating a bunch of views that are never used.

FunctionalTableData/CellStyle.swift
```diff
@@ -39,15 +39,16 @@ public struct CellStyle {
 	public var selectionColor: UIColor? = CellStyle.selectionColor
 	/// The view's background color.
 	public var backgroundColor: UIColor?
-	/// The view that is displayed behind the cell’s other content.
-	public var backgroundView: UIView?
```
krbarnes: This is a breaking API change since we're deleting a public property. The fix is pretty trivial. Should we keep the property and maintain backward compatibility, or just break clients and increment the major version number?
@raulriera @g-Off

FunctionalTableData/CellStyle.swift
```diff
@@ -39,15 +39,16 @@ public struct CellStyle {
...
+	/// Provides the view that is displayed behind the cell's other content.
+	public var backgroundViewProvider: BackgroundViewProvider
 	/// The tint color to apply to the cell.
 	public var tintColor: UIColor?
 	/// The default spacing to use when laying out content in the view.
 	public var layoutMargins: UIEdgeInsets?
 	/// The radius to use when drawing rounded corners in the view.
 	public var cornerRadius: CGFloat
 	
+	@available(*, unavailable, message: "The `backgroundView` argument is no longer available. Use backgroundViewProvider instead.")
```
g-Off: should mark as deprecated and not unavailable

FunctionalTableData/CellStyle.swift
g-Off: can make this a computed property, getter returns result of backgroundViewProvider and setter wraps and creates a backgroundViewProvider

raulriera: We should keep the property, its already using a second initializer. We can just deprecate this property and fix all the usages. That is what we have been doing lately

FunctionalTableData/CellStyle.swift
```diff
@@ -160,10 +185,17 @@ extension CellStyle: Equatable {
 		equality = equality && lhs.accessoryType == rhs.accessoryType
 		equality = equality && lhs.selectionColor == rhs.selectionColor
 		equality = equality && lhs.backgroundColor == rhs.backgroundColor
-		equality = equality && lhs.backgroundView == rhs.backgroundView
```
g-Off: I think we need to expand this change a bit.
backgroundViewProvider should be used as part of the equality check (I think) because you can have the same cell key and state but provide a different backgroundViewProvider which should force it to refresh that cell.

I'm not sure how we make BackgroundViewProvider equatable though…

FunctionalTableData/CellStyle.swift
Campbell-Scott: Great, i'll make this change

FunctionalTableData/CellStyle.swift
dannbeaus: give it an ID? but how not to clash 🤔

FunctionalTableData/CellStyle.swift
dannbeaus: setter? Should it not be deferred in a closure?

FunctionalTableData/CellStyle.swift
krbarnes: We could force some `state:Equatable` that the user implements. 
The other problem with this approach is the background view is still associated to the ViewState instead of the cell and this makes reuse impossible. If we're scrolling through a long list and all cells have a backgroundView providers, those views will be initialized on the fly and not reused even if the cell types are the same.

FunctionalTableData/CellStyle.swift
Campbell-Scott: When I mark this deprecated over unavailable I get `ambiguous use of init` wherever we init as `CellStyle()`. Any idea how this can be avoided?

FunctionalTableData/CellStyle.swift
```diff
@@ -39,23 +39,27 @@ public struct CellStyle {
 	public var selectionColor: UIColor? = CellStyle.selectionColor
 	/// The view's background color.
 	public var backgroundColor: UIColor?
-	/// The view that is displayed behind the cell’s other content.
+	/// The view that is displayed behind the cell's other content.
+	@available(*, deprecated, message: "The `backgroundView` property is no longer available. User backgroundViewProvider instead.")
```
krbarnes: I think this can be just "Replaced with backgroundViewProvider"

dannbeaus:
I like @krbarnes suggestion, it fits the requirement for being evaluated as equatable with out having to create the view

FunctionalTableData/BackgroundViewProvider.swift
```diff
@@ -0,0 +1,34 @@
...
+	func backgroundView() -> UIView?
+
+	/// Compare the internal state to avoid unnecessarily instantiating the background view.
+	func isEqualTo(_ other: BackgroundViewProvider) -> Bool
+}
+
+public extension BackgroundViewProvider {
+	public func isEqualTo(_ other: BackgroundViewProvider) -> Bool {
+		return type(of: self) == type(of: other)
+	}
```
Caiopia: I'm thinking that if we provide a default implementation it might not be obvious for implementers of `BackgroundViewProvider` to realize that they need more than a just type check. Are we able to omit this to force implementers to provide their own `isEquals()`?

Campbell-Scott: I would agree with you here.

FunctionalTableData/CellStyle.swift
```diff
@@ -54,23 +54,27 @@ public struct CellStyle {
...
+	@available(*, deprecated, message: "The `backgroundView` argument is no longer available. Use backgroundViewProvider instead.")
 	public init(topSeparator: Separator.Style? = nil,
 	            bottomSeparator: Separator.Style? = nil,
 	            separatorColor: UIColor? = nil,
 	            highlight: Bool? = nil,
 	            accessoryType: UITableViewCell.AccessoryType = .none,
 	            selectionColor: UIColor? = CellStyle.defaultSelectionColor,
 	            backgroundColor: UIColor? = CellStyle.defaultBackgroundColor,
-	            backgroundView: UIView? = nil,
+	            backgroundView: UIView?,
```
Caiopia: We should probably keep the same default parameter values, no?

Issue #113, opened 2018-10-09
Table view frame accessed from background thread


stevenwenxu:
`tableView.frame` is accessed directly in the `BlockOperation`, this is caught by the main thread checker.

<img width="1179" alt="screen shot 2018-10-09 at 1 32 06 pm" src="https://user-images.githubusercontent.com/3827809/46687940-263a7e00-cbca-11e8-9bbf-209b020a7a69.png">

dannbeaus:
do you have your stack trace available?
how did you reproduce in your scenario?

stevenwenxu:
It is a crash in POS where we have multiple rows with the same keys. However this isn't specific to any case because by looking at the code, the failure block is always run in the `blockOperation` which is on the `renderAndDiffQueue`.

PR #114, opened 2018-10-09
Fixed background thread access UI component

stevenwenxu:
Resolves https://github.com/Shopify/FunctionalTableData/issues/113

The `tableView.frame` was directly accessed in the `BlockOperation`. As suggested by the main thread checker, this should only be accessed in the main thread.

### DIFF
--- a/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
+++ b/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
@@ -202,7 +202,8 @@ public class FunctionalCollectionData: NSObject {
 					if $0.name == NSExceptionName.internalInconsistencyException {
 						guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
 						let changes = TableSectionChangeSet()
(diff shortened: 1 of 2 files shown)


FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -202,7 +202,11 @@ public class FunctionalCollectionData: NSObject {
 					if $0.name == NSExceptionName.internalInconsistencyException {
 						guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
 						let changes = TableSectionChangeSet()
-						let exception = FunctionalTableData.Exception(name: $0.name.rawValue, newSections: newSections, oldSections: strongSelf.sections, changes: changes, visible: [], viewFrame: strongSelf.collectionView?.frame ?? .zero, reason: $0.reason, userInfo: $0.userInfo)
+						var viewFrame: CGRect = .zero
```
g-Off: I think you can write this as:
`let viewFrame = DispatchQueue.main.sync { return strongSelf.collectionView?.frame ?? .zero }`

dannbeaus:
> I might capture the frame before starting the operation

I don't think the public func `renderAndDiff` is guaranteed to be called on the main queue; at least with the block operation, its guaranteed to be not the main queue, so `sync` on main queue should be safe from here.

PR #117, opened 2018-10-23
Use performBatchUpdates on iOS11+

krbarnes:
Trying to track down some crashes that look like this.
```
"exception" : "attempt to delete row 0 from section 3, but there are only 3 sections before the update"
```

Happens somewhat irregularly and I haven't been able to find a way to reproduce it. Was curious is using the new `performBatchUpdates` API would fix it.

Tests are still passing on iOS 10 and 12, but I'm not sure this PR is worth merging.

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -400,18 +400,31 @@ public class FunctionalTableData: NSObject {
 		CATransaction.setCompletionBlock {
 			completion?()
 		}
-		
-		tableView.beginUpdates()
+
 		// #4629 - There is an issue where on some occasions calling beginUpdates() will cause a heightForRowAtIndexPath() call to be made. If the sections have been changed already we may no longer find the cells
 		// in the model causing a crash. To prevent this from happening, only load the new model AFTER beginUpdates() has run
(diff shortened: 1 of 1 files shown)


FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -400,18 +400,31 @@ public class FunctionalTableData: NSObject {
...
 		}
-		
-		tableView.beginUpdates()
+
 		// #4629 - There is an issue where on some occasions calling beginUpdates() will cause a heightForRowAtIndexPath() call to be made. If the sections have been changed already we may no longer find the cells
 		// in the model causing a crash. To prevent this from happening, only load the new model AFTER beginUpdates() has run
-		sections = localSections
-		applyTableSectionChanges(changes)
-		tableView.endUpdates()
+		if #available(iOS 11.0, *) {
```
Caiopia: Could be worth separating these branches into 2 inline functions so we can handle the completion handler differently for both and don't have as many `if #available` checks.

Does the `performBatchUpdates` need to be within the `CATransaction`s?

raulriera: I am with you on this Caio

Issue #105, opened 2018-07-30
Deprecate (and remove?) CellStyle.backgroundView


krbarnes:
The `CellStyle.backgroundView` property is too expensive. Views should only be created when they're needed. This background view property means that the view is created and destroyed on every render pass even though it's rarely used.

PR #123, opened 2018-12-10
Deprecate backgroundView for backgroundViewProvider

Campbell-Scott:
Re-opening this as changes have been pushed to the old branch since `stalebot` killed the PR.

The `backgroundView` property on `CellStyle` is expensive - background views are created on every render pass, even though they are rarely used. 

The `BackgroundViewProvider` allows `CellStyle` to request an optional background view when its being dequeued; the only time it's actually needed. 

Fixes https://github.com/Shopify/FunctionalTableData/issues/105

### DIFF
--- a/FunctionalTableData/BackgroundViewProvider.swift
+++ b/FunctionalTableData/BackgroundViewProvider.swift
@@ -1,0 +1,20 @@
+//
+//  BackgroundViewProvider.swift
+//  FunctionalTableData
+//
+//  Created by Scott Campbell on 2018-10-16.
+//  Copyright © 2018 Shopify. All rights reserved.
+//
+
+import UIKit
+
+/// Provide a background view to be displayed behind the other contents of a cell.
+/// An implementation should maintain some internal state about the contents of the view.
+public protocol BackgroundViewProvider {
+	/// This is where the background view should be instantiated, since this
+	/// function is only called when a cell is being prepared to be dequeued.
+	func backgroundView() -> UIView?
+
+	/// Compare the internal state to avoid unnecessarily instantiating the background view.
+	func isEqualTo(_ other: BackgroundViewProvider?) -> Bool
+}
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -54,8 +54,11 @@ public struct CellStyle {
 	public var selectionColor: UIColor?
 	/// The view's background color.
 	public var backgroundColor: UIColor?
-	/// The view that is displayed behind the cell’s other content.
+	/// The view that is displayed behind the cell's other content.
+	@available(*, deprecated, message: "Replaced with backgroundViewProvider.")
 	public var backgroundView: UIView?
+	/// Provides the view that is displayed behind the cell's other content.
+	public var backgroundViewProvider: BackgroundViewProvider?
 	/// The tint color to apply to the cell.
 	public var tintColor: UIColor?
 	/// The default spacing to use when laying out content in the view.
@@ -63,6 +66,7 @@ public struct CellStyle {
 	/// The radius to use when drawing rounded corners in the view.
 	public var cornerRadius: CGFloat
 	
+	@available(*, deprecated, message: "The `backgroundView` argument is no longer available. Use backgroundViewProvider instead.")
 	public init(topSeparator: Separator.Style? = nil,
 	            bottomSeparator: Separator.Style? = nil,
 	            separatorColor: UIColor? = nil,
@@ -70,7 +74,7 @@ public struct CellStyle {
 	            accessoryType: UITableViewCell.AccessoryType = .none,
 	            selectionColor: UIColor? = CellStyle.defaultSelectionColor,
 	            backgroundColor: UIColor? = CellStyle.defaultBackgroundColor,
-	            backgroundView: UIView? = nil,
+	            backgroundView: UIView?,
 	            tintColor: UIColor? = nil,
 	            layoutMargins: UIEdgeInsets? = nil,
 	            cornerRadius: CGFloat = 0) {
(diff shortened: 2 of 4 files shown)


FunctionalTableData/CellStyle.swift
```diff
@@ -81,16 +85,59 @@ public struct CellStyle {
 		self.accessoryType = accessoryType
 		self.selectionColor = selectionColor
 		self.backgroundColor = backgroundColor
-		self.backgroundView = backgroundView
+		self.tintColor = tintColor
+		self.layoutMargins = layoutMargins
+		self.cornerRadius = cornerRadius
+
+		struct DefaultBackgroundProvider: BackgroundViewProvider {
```
Campbell-Scott: The `DefaultBackgroundProvider` in the deprecated init simply wraps and returns the (optionally) provided `backgroundView`. So we'll still get the functionality in this case.

FunctionalTableData/CellStyle.swift
```diff
@@ -54,23 +54,27 @@ public struct CellStyle {
...
+	@available(*, deprecated, message: "The `backgroundView` argument is no longer available. Use backgroundViewProvider instead.")
 	public init(topSeparator: Separator.Style? = nil,
 	            bottomSeparator: Separator.Style? = nil,
 	            separatorColor: UIColor? = nil,
 	            highlight: Bool? = nil,
 	            accessoryType: UITableViewCell.AccessoryType = .none,
 	            selectionColor: UIColor? = CellStyle.defaultSelectionColor,
 	            backgroundColor: UIColor? = CellStyle.defaultBackgroundColor,
-	            backgroundView: UIView? = nil,
+	            backgroundView: UIView?,
```
Campbell-Scott: Removed the default value here to avoid `Ambiguous use of 'init'`. This also means `CellStyle` initializers without an explicitly set `backgroundView` will default to the new implementation.

FunctionalTableDataTests/CellStyleTests.swift
```diff
@@ -13,6 +13,21 @@ class StyleTests: XCTestCase {
...
 	var style: CellStyle!
+
+	struct ColoredBackgroundProvider: BackgroundViewProvider {
+		let color: UIColor
+
+		public func backgroundView() -> UIView? {
+			let bgView = UIView()
+			bgView.backgroundColor = color
+			return bgView
+		}
```
raulriera: This will still create a new view every time we call this method. I thought this is what we are trying to avoid

FunctionalTableData/BackgroundViewProvider.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+
+import UIKit
+
+/// Provide a background view to be displayed behind the other contents of a cell.
+/// An implementation should maintain some internal state about the contents of the view.
+public protocol BackgroundViewProvider {
+	/// This is where the background view should be instantiated, since this
+	/// function is only called when a cell is being prepared to be deqeueued.
+	func backgroundView() -> UIView?
```
raulriera: Shouldn't this be a property instead to avoid the problem mention above? Or at least make it "harder to bypass"?

FunctionalTableDataTests/CellStyleTests.swift
Campbell-Scott: `CellStyle.configure(...)` calls this method, which is the correct time to create and apply the view to the cell. So instead of this:
```swift
let backgroundView = someView(...)
let cellStyle = CellStyle(backgroundView: backgroundView)
```
we can create a provider, which knows how to create the view, and init with that instead. Which would look like:

```swift
struct YellowProvider: BackgroundViewProvider {
        let backgroundColor: UIColor

        func backgroundView() -> UIView? {
                let view = UIView()
                view.backgroundColor = backgroundColor
                return view
        }
        ...
}

let provider = YellowProvider(backgroundColor: .red)
let cellStyle = CellStyle(backgroundViewProvider: provider)
```

This way we don't create the view in the builder (stopping ourselves from creating a view even if we're just to doing a comparison, not actually drawing anything), but instead we create it in `CellStye.configure(...)`.

FunctionalTableData/BackgroundViewProvider.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Copyright © 2018 Shopify. All rights reserved.
+//
+
+import UIKit
+
+/// Provide a background view to be displayed behind the other contents of a cell.
+/// An implementation should maintain some internal state about the contents of the view.
+public protocol BackgroundViewProvider {
+	/// This is where the background view should be instantiated, since this
+	/// function is only called when a cell is being prepared to be deqeueued.
```
raulriera: Small typo in the last word

raulriera:
The documentation should have been another PR 😬 sorry for not clarifying that

Campbell-Scott:
@raulriera No worries - i'll separate them 👍

FunctionalTableData/CellStyle.swift
```diff
@@ -163,10 +204,10 @@ extension CellStyle: Equatable {
 		equality = equality && lhs.accessoryType == rhs.accessoryType
 		equality = equality && lhs.selectionColor == rhs.selectionColor
 		equality = equality && lhs.backgroundColor == rhs.backgroundColor
-		equality = equality && lhs.backgroundView == rhs.backgroundView
 		equality = equality && lhs.tintColor == rhs.tintColor
 		equality = equality && lhs.layoutMargins == rhs.layoutMargins
 		equality = equality && lhs.cornerRadius == rhs.cornerRadius
+		equality = equality && lhs.backgroundViewProvider?.isEqualTo(rhs.backgroundViewProvider) ?? true
```
josephroqueca: What if `lhs.backgroundViewProvider` is `nil` and `rhs.backgroundViewProvider` is not?

```suggestion
		equality = equality && lhs.backgroundViewProvider?.isEqualTo(rhs.backgroundViewProvider) ?? ((lhs.backgroundViewProvider == nil) == (rhs.backgroundViewProvider == nil))
```

Campbell-Scott: I think in this case it is sufficient to write the equability check as:
```
lhs.backgroundViewProvider?.isEqualTo(rhs.backgroundViewProvider) ?? (rhs.backgroundViewProvider == nil)
```

PR #125, opened 2018-12-12
Generate jazzy documentation.

Campbell-Scott:
Ran `jazzy` to update the docs for the new `BackgroundViewProvider`.


docs/undocumented.json
```diff
@@ -1,978 +1,1020 @@
 {
   "warnings": [
     {
-      "file": "/Users/raulriera/Downloads/Shopify/FunctionalTableData/FunctionalTableData/CellActions.swift",
+      "file": "/Users/scottcampbell/src/github.com/Shopify/FunctionalTableData/FunctionalTableData/CellActions.swift",
```
raulriera: Maybe this file shouldn't have been committed 🤔 but that was on me

Campbell-Scott: Looks like I can just remove the `undocumented.json` 🤷‍♂️

raulriera: Kill it, and maybe gitignore? 🤔

Campbell-Scott: 👍

PR #129, opened 2019-01-10
Include section header state changes in sections diff

Campbell-Scott:
**Problem**
The state of a `TableSection`s header/footer is not considered in the tableView diff. This means that changes to a header or footer's state will not cause its `TableSection` to be reloaded during a `renderAndDiff`.

 **Solution**
Add `isEqual(...) -> Bool` to the `TableHeaderFooterConfigType` protocol. This allows our conforming types to define what equality means. There is also an extension with a default implementation so that these changes are not breaking - simply check to make sure the types match.

A header/footers updater is only called when a header/footer is dequeued. The only way to ensure a header/footer is dequeued is to reload the table section it belongs to. So, when building the change set:
1. If the section has been deleted, inserted, or moved, this case is irrelevant (the section will go away, or its contents will be dequeued).
2. If the section is in the same position the logic becomes: 
```
if footer has changed or header has changed {
    // The header/footer has changed so we need to reload the whole section in order to update its view. 
    // Skip building the rows changeSet, because all rows will be updated anyways.
} else {
    // The header/footers state has not changed, so compareRows(...) as normal.
}
```

**Potential drawbacks**
Case: We have a section with an arbitrarily large number of rows. There are no state changes for any of the rows, but the header's state has changed. In this instance, reload the entire section in order to update the headers view. I still think this is the appropriate action, as the header is a defining characteristic of the section; if the header changes, we would expect the section to be reloaded.

**Example of implementation** 
In this example, we can construct a section header/footer with a view and state that is not purely decorative. Changes to the headers/footers state will cause the section to reload, so the view is updated accordingly.

```swift
struct CustomHeaderState: TableHeaderFooterStateType, Equatable {
        let insets: UIEdgeInsets
        let topSeparatorHidden: Bool
        let bottomSeparatorHidden: Bool
        let height: CGFloat

        // Any additional stuff thats important for your custom state
        let foo: Int
        let bar: Int
}

struct CustomHeaderFooter<ViewType: UIView, Layout: TableItemLayout>: TableHeaderFooterConfigType {
        typealias ViewUpdater = (_ header: TableHeaderFooter<ViewType, Layout>, _ state: CustomHeaderState) -> Void
        let state: CustomHeaderState?
        let updater: ViewUpdater?

        init(state: CustomHeaderState? = nil, updater: ViewUpdater? = nil) {
                self.state = state
                self.updater = updater
        }

       func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool {
                guard let other = other as? CustomHeaderFooter else { return false }
                return state == other.state
        }

        ...
}
```

### DIFF
--- a/FunctionalTableData/TableSectionChangeSet.swift
+++ b/FunctionalTableData/TableSectionChangeSet.swift
@@ -134,7 +134,11 @@ public final class TableSectionChangeSet {
 				if oldSectionIndex < old.count && new[newSectionIndex].key == old[oldSectionIndex].key {
 					// Skip over equal sections
 					repeat {
-						compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
+						if headerOrFooterChanged(oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex) {
+							reloadedSections.insert(newSectionIndex)
+						} else {
+							compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
+						}
 						oldSectionIndex += 1
 						newSectionIndex += 1
 					} while oldSectionIndex < old.count &&
@@ -165,6 +169,22 @@ public final class TableSectionChangeSet {
 			return false
 		}
 		return new.section.mergedStyle(for: new.row) == old.section.mergedStyle(for: old.row)
+	}
+
+	private func headerOrFooterChanged(oldSectionIndex: Int, newSectionIndex: Int) -> Bool {
+		let oldHeader = old[oldSectionIndex].header
+		let newHeader = new[newSectionIndex].header
+		guard oldHeader?.isEqual(newHeader) ?? (newHeader == nil) else {
+			return true
+		}
+
+		let oldFooter = old[oldSectionIndex].footer
+		let newFooter = new[newSectionIndex].footer
+		guard oldFooter?.isEqual(newFooter) ?? (newFooter == nil) else {
+			return true
+		}
+
+		return false
 	}
 
 	private func compareRows(newRows: inout Set<String>, oldRows: inout [String: Int], oldSectionIndex: Int, newSectionIndex: Int) {
--- a/FunctionalTableData/TableSectionHeaderFooter.swift
+++ b/FunctionalTableData/TableSectionHeaderFooter.swift
@@ -15,7 +15,14 @@ public typealias TableHeaderConfigType = TableHeaderFooterConfigType
 
 public protocol TableHeaderFooterConfigType: TableItemConfigType {
 	func dequeueHeaderFooter(from tableView: UITableView) -> UITableViewHeaderFooterView?
+	func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool
 	var height: CGFloat { get }
+}
+
+public extension TableHeaderFooterConfigType {
+	func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool {
+		return (other as? Self) != nil
+	}
 }
 
 public protocol TableHeaderFooterStateType {
--- a/FunctionalTableDataTests/TableSectionChangeSetTests.swift
+++ b/FunctionalTableDataTests/TableSectionChangeSetTests.swift
@@ -63,6 +63,90 @@ class TableSectionChangeSetTests: XCTestCase {
 		XCTAssertEqual(changes.deletedSections, IndexSet())
 	}
 
+	func testSectionNotReloadedWithEqualHeaders() {
+		let oldItems: [TableSection] = [TableSection(key: "section1", header: TestHeaderFooter(state: TestHeaderFooterState(data: "green")))]
+		let newItems: [TableSection] = [TableSection(key: "section1", header: TestHeaderFooter(state: TestHeaderFooterState(data: "green")))]
+		let changes = TableSectionChangeSet(old: oldItems, new: newItems, visibleIndexPaths: [])
+
+		XCTAssertTrue(changes.isEmpty)
+		XCTAssertEqual(changes.movedSections, [])
+		XCTAssertEqual(changes.insertedSections, IndexSet())
+		XCTAssertEqual(changes.reloadedSections, IndexSet())
+		XCTAssertEqual(changes.deletedSections, IndexSet())
+	}
+
+	func testSectionNotReloadedWithEqualFooters() {
+		let oldItems: [TableSection] = [TableSection(key: "section1", footer: TestHeaderFooter(state: TestHeaderFooterState(data: "blue")))]
+		let newItems: [TableSection] = [TableSection(key: "section1", footer: TestHeaderFooter(state: TestHeaderFooterState(data: "blue")))]
+		let changes = TableSectionChangeSet(old: oldItems, new: newItems, visibleIndexPaths: [])
+
+		XCTAssertTrue(changes.isEmpty)
+		XCTAssertEqual(changes.movedSections, [])
+		XCTAssertEqual(changes.insertedSections, IndexSet())
+		XCTAssertEqual(changes.reloadedSections, IndexSet())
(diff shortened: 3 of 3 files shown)


FunctionalTableData/TableSectionHeaderFooter.swift
```diff
@@ -15,9 +15,16 @@ public typealias TableHeaderConfigType = TableHeaderFooterConfigType
 
 public protocol TableHeaderFooterConfigType: TableItemConfigType {
 	func dequeueHeaderFooter(from tableView: UITableView) -> UITableViewHeaderFooterView?
+	func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool
 	var height: CGFloat { get }
 }
 
+public extension TableHeaderFooterConfigType {
+	func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool {
+		return (other as? Self) != nil
```
josephroqueca: Why does this have to be cast to `Self` if the parameter is already constrained to an identical type? Can this just be:

```suggestion
		return other != nil
```

Campbell-Scott: No, if we had the following:
```
struct Header1: TableSectionHeaderConfigType {}
struct Header2: TableSectionHeaderConfigType {}
```

 at runtime the type of `Self` would evaluate to one of `Header1` or `Header2`, but the type of `other` evaluates to `TableSectionHeaderConfigType?`. We need to cast `other` to `Self` to ensure that we have two instances of `Header1`, and not a instance of `Header1` and an instance of `Header2`.

FunctionalTableData/TableSectionChangeSet.swift
```diff
@@ -134,7 +134,11 @@ public final class TableSectionChangeSet {
 				if oldSectionIndex < old.count && new[newSectionIndex].key == old[oldSectionIndex].key {
 					// Skip over equal sections
 					repeat {
-						compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
+						if old[oldSectionIndex] != new[newSectionIndex] {
```
Caiopia: Can we add a test case for this in `TableSectionChangeSetTests`?

FunctionalTableData/TableSectionChangeSet.swift
```diff
@@ -134,7 +134,11 @@ public final class TableSectionChangeSet {
...
+							reloadedSections.insert(newSectionIndex)
+						} else {
+							compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
+						}
```
raulriera: Wait? why do we need this? by changing `extension TableSection: Equatable` changes in the header will be taken into effect right? Another solution would be to no do the `extension TableSection: Equatable` but add a `compareHeader` similar to what we have here in `compareRows`.

But lastly, whatever you do in the headers needs to be done for the footers

FunctionalTableDataTests/TableSectionChangeSetTests.swift
```diff
@@ -63,6 +63,31 @@ class TableSectionChangeSetTests: XCTestCase {
 		XCTAssertEqual(changes.deletedSections, IndexSet())
 	}
 
+	func testSectionEquality() {
```
raulriera: Be more expressive here `testNothingIsDiffedWithEqualSections` or something like that

FunctionalTableData/TableSectionHeaderFooter.swift
raulriera: What is happening here, this is not checking or equality

Campbell-Scott: I didn't want to just add `isEqual(...)` to the protocol without a default implementation as it would become a breaking change; classes no longer conforming to it. The default is only checking that the headers are of the same type.

FunctionalTableData/TableSectionChangeSet.swift
Campbell-Scott: No, we can make the changes to `extension TableSection: Equatable`, but there is never anything that checks the section equality that will trigger a reload for that section. We need to add the section index to `reloadedSections` in order to see the changes take effect. 

A `compareHeader` would read better I think 👍
I can implement the changes for footers as well.

Campbell-Scott: But this can be simplified as you suggested, so I'll create a `compareHeaders()` or something, and just compare the states there, so that we don't need to touch the sections equatable at all.

FunctionalTableData/TableSectionChangeSet.swift
```diff
@@ -134,7 +134,11 @@ public final class TableSectionChangeSet {
 				if oldSectionIndex < old.count && new[newSectionIndex].key == old[oldSectionIndex].key {
 					// Skip over equal sections
 					repeat {
-						compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
+						if headerOrFooterChanged(oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex) {
+							reloadedSections.insert(newSectionIndex)
+						} else {
+							compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
```
raulriera: I believe this should be

```swift
compareSectionsHeadersFooters()
compareRows()
```

otherwise if they are equal the rows are entirely skipped right?

Campbell-Scott: `headerOrFooterChanged(...)` returns true if the state of the header or the state of the footer has changed. In either of those cases, we need to reload the whole section to update the appropriate view, so we can avoid comparing rows. 
We only need to compare rows if `headerOrFooterChanged(...)` returns false.

FunctionalTableData/TableSectionHeaderFooter.swift
raulriera: We should still fix this everywhere we have it (and maybe explain to people the impact of putting dynamic things in the header and footer)

FunctionalTableDataTests/TableSectionChangeSetTests.swift
```diff
@@ -63,6 +63,31 @@ class TableSectionChangeSetTests: XCTestCase {
...
+		let oldItems: [TableSection] = [TableSection(key: "section1", header: TestHeader(state: TestHeaderState(data: "state1")))]
+		let newItems: [TableSection] = [TableSection(key: "section1", header: TestHeader(state: TestHeaderState(data: "state1")))]
+		let changes = TableSectionChangeSet(old: oldItems, new: newItems, visibleIndexPaths: [])
+
+		XCTAssertTrue(changes.isEmpty)
+		XCTAssertEqual(changes.movedSections, [])
+		XCTAssertEqual(changes.insertedSections, IndexSet())
+		XCTAssertEqual(changes.reloadedSections, IndexSet())
+		XCTAssertEqual(changes.deletedSections, IndexSet())
+	}
```
raulriera: Can you also add rows to oldItems and newItems? So we can check the `else` when the headers are the same

FunctionalTableDataTests/TableSectionChangeSetTests.swift
```diff
@@ -63,6 +63,31 @@ class TableSectionChangeSetTests: XCTestCase {
...
+		let changes = TableSectionChangeSet(old: oldItems, new: newItems, visibleIndexPaths: [])
+
+		XCTAssertFalse(changes.isEmpty)
+		XCTAssertEqual(changes.count, 1)
+		XCTAssertEqual(changes.movedSections, [])
+		XCTAssertEqual(changes.insertedSections, IndexSet())
+		XCTAssertEqual(changes.reloadedSections, IndexSet([0]))
+		XCTAssertEqual(changes.deletedSections, IndexSet())
+	}
+
```
raulriera: we are missing footers tests 🙂

Campbell-Scott: Tests updated to add rows, and footer tests added

PR #130, opened 2019-01-22
BUG FIX: Inserting wrong index into reloadedSections.

Campbell-Scott:
### Issue: 
The wrong index was inserted into `reloadedSections` in the `TableSectionChangeSet`. This should have been `oldSectionIndex` and not `newSectionIndex` 😓

### Errors:
1. `NSInternalInconsistencyException: Attempt to delete and reload the same index path`
```
OLD [ A, B, C, D ]  // Delete B; Reload C
NEW [ A, C, D ]

Expected change set:
deletedSections = [ 1 ]
reloadedSections = [ 2 ]

Actual change set:
deletedSections = [ 1 ]
reloadedSections = [ 1 ]
```

2. `NSInternalInconsistencyException: Attempt to delete section x, but there are only x sections before the update `
```
OLD = [ A ]  // Insert B; Reload A
NEW = [ B, A ]

Expected change set: 
insertedSections = [ 0 ]
reloadedSections = [ 0 ]

Actual change set:
insertedSections = [ 0 ]
reloadedSections = [ 1 ]  // OLD[1] doesn't exist.
```

## Fix:
- Insert `oldSectionIndex` into `reloadedSections` instead of `newSectionIndex`.
- Unit tests that cover the two cases outlined above. When `newSectionIndex` is used, both of these tests fail.

### DIFF
--- a/FunctionalTableData/TableSectionChangeSet.swift
+++ b/FunctionalTableData/TableSectionChangeSet.swift
@@ -135,7 +135,7 @@ public final class TableSectionChangeSet {
 					// Skip over equal sections
 					repeat {
 						if headerOrFooterChanged(oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex) {
-							reloadedSections.insert(newSectionIndex)
+							reloadedSections.insert(oldSectionIndex)
 						} else {
 							compareRows(newRows: &newRows, oldRows: &oldRows, oldSectionIndex: oldSectionIndex, newSectionIndex: newSectionIndex)
 						}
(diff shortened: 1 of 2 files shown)


Campbell-Scott:
The existing tests gave a false positive, ex:
```
OLD = [ A, B ]  // Reload B
NEW = [ A, B ]

Expected: 
reloadedSections = [ 1 ]

Actual:
reloadedSections = [ 1 ]
```
Which _appeared_ correct, but if we insert a section at OLD[ 1 ] we get:
```
Expected: 
insertedSections = [ 1 ]
reloadedSections = [ 2 ]

Actual:
insertedSections = [ 1 ]
reloadedSections = [ 1 ]
```
which is incorrect.

PR #132, opened 2019-02-05
TableSectionHeaderFooter / HostCell parity

Campbell-Scott:
### Proposed Changes
Get `TableSectionHeaderFooter` closer to our `HostCell`.
1. Make `TableHeaderFooterStateType` equatable:   
    - Implement a custom `TableSectionHeaderFooter` with a view, state, and layout in the same way you would a `HostCell`; states compared in the same way.   
    - Remove the need for custom `TableHeaderFooterConfigType`s that only exist to provide custom `isEqual` implementations for non-default headers and footers.   
    - An example of this is the additions/deletions for `TableSectionChangeSetTests.swift`.
2. Add `debugInfo` too `TableHeaderFooterConfigType`:   
    - Carried over from `HostCell` to provide more descriptive logging when debugging state diffs.

### Breaking
Because of protocol additions, these changes are breaking for those who have implemented custom `TableHeaderFooterConfig`s  outside of the default `TableSectionHeaderFooter` that we provide.

### DIFF
--- a/FunctionalTableData/TableSectionHeaderFooter.swift
+++ b/FunctionalTableData/TableSectionHeaderFooter.swift
@@ -14,9 +14,10 @@ import UIKit
 public typealias TableHeaderConfigType = TableHeaderFooterConfigType
 
 public protocol TableHeaderFooterConfigType: TableItemConfigType {
+	var height: CGFloat { get }
 	func dequeueHeaderFooter(from tableView: UITableView) -> UITableViewHeaderFooterView?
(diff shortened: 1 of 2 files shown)


Campbell-Scott:
@raulriera Small changes: remove default implementation of `isEqual`, make `TableHeaderFooterStateType` equatable, add `debugInfo` to `TableHeaderFooterConfigType`.

Caiopia:
We talked offline about a way to make this a non-breaking change so that we don't accidentally create a breaking change in a future patch release. Could we update this PR with that change?

PR #134, opened 2019-02-27
Added where into for loop 

alexliubj:
1.  Added `where` into second for loop to make code cleaner
2. Optimized `duplicates()` method in Array extension using `Dictionary(grouping:)` and filter function .

### DIFF
--- a/FunctionalTableData/Extensions/Array+TableSection.swift
+++ b/FunctionalTableData/Extensions/Array+TableSection.swift
@@ -29,18 +29,10 @@ extension Array where Element: TableSectionType {
 }
 
 private extension Array where Element: Hashable {
-	func duplicates() -> [Element] {
-		var dups: Set<Element> = []
-		var uniques: Set<Element> = []
-		forEach {
-			if uniques.contains($0) {
-				dups.insert($0)
-			} else {
-				uniques.insert($0)
-			}
-		}
-		return Array(dups)
-	}
+    func duplicates() -> [Element] {
+        let groups = Dictionary(grouping: self, by: { $0 }).filter{ $1.count > 1 }
+        return Array(groups.keys)
+    }
 }
 
 private extension Array where Element: TableSectionType {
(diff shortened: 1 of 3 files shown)


robinsenior:
If I'm not mistaken, neither `keyPathForRowKey` nor `sectionForKey` are actually referenced and could be deleted entirely.

alexliubj:
@robinsenior, looks like not being referenced so far in the entire project.

alexliubj:
Btw, how can I manually trigger bot again after I sign CLA, it says will do it automatically but didn't happen.

robinsenior:
@alexliubj I've done it for you, but you need to click on the failed check and ask it to re-run (maybe you didn't have permissions? 🤔)

robinsenior:
@alexliubj if you want to delete the dead code I'll get some eyes on this

alexliubj:
@robinsenior Sure, will do it later today. `sectionForKey` is not referenced and it is a private function. `
keyPathForRowKey` is public and there is a test case for it, I think this method should stay from user perspective.

alexliubj:
@robinsenior removed `sectionForKey(key: String)` and please check.

Caiopia:
Good call on `keyPathForRowKey`. LGTM

FunctionalTableData/Extensions/Array+TableSection.swift
```diff
@@ -29,18 +29,10 @@ extension Array where Element: TableSectionType {
...
-			} else {
-				uniques.insert($0)
-			}
-		}
-		return Array(dups)
-	}
+    func duplicates() -> [Element] {
+        let groups = Dictionary(grouping: self, by: { $0 }).filter{ $1.count > 1 }
+        return Array(groups.keys)
+    }
```
Caiopia: Oh, just noticed this this is using spaces instead of tabs. Can we update it to be like the rest of the file?

alexliubj: @Caiopia After I tried a couple times, I believe this is fixed.

Caiopia:
Awesome, thanks @alexliubj. Going to merge this.

PR #137, opened 2019-03-08
Adds support for setting a UIScrollViewDelegate

g-Off:
This is for use instead of the closure based mechanism

Uses NSObject's forwardingTarget to send UIScrollViewDelegate based messages to a separate object and removes the need for Functional(Table|Collection)Data to implement and forward all of the various UIScrollViewDelegate based messages itself. Backwards compatibility is maintained with a separate internal object that maintains the closures and will call them (so long as a delegate implementing the same method doesn't exist). In future we can remove this backwards compatible support and rely solely on the UIScrollViewDelegate.

### DIFF
--- a/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
+++ b/FunctionalTableData/CollectionView/FunctionalCollectionData.swift
@@ -64,30 +64,39 @@ public class FunctionalCollectionData: NSObject {
 		return sections[indexPath]
 	}
 	
-	/// See UIScrollView's [documentation](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619392-scrollviewdidscroll) for more information.
-	public var scrollViewDidScroll: ((_ scrollView: UIScrollView) -> Void)?
-	/// See UIScrollView's [documentation](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619394-scrollviewwillbegindragging) for more information.
-	public var scrollViewWillBeginDragging: ((_ scrollView: UIScrollView) -> Void)?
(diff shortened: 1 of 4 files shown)


FunctionalTableData/CollectionView/FunctionalCollectionData.swift
```diff
@@ -64,30 +64,38 @@ public class FunctionalCollectionData: NSObject {
...
-	public var scrollViewDidEndScrollingAnimation: ((_ scrollView: UIScrollView) -> Void)?
-	/// See UIScrollView's [documentation](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619378-scrollviewshouldscrolltotop) for more information.
-	public var scrollViewShouldScrollToTop: ((_ scrollView: UIScrollView) -> Bool)?
-	/// See UIScrollView's [documentation](https://developer.apple.com/documentation/uikit/uiscrollviewdelegate/1619382-scrollviewdidscrolltotop) for more information.
-	public var scrollViewDidScrollToTop: ((_ scrollView: UIScrollView) -> Void)?
-	
-	/// An optional callback that describes the current scroll position of the collection view as an accessibility aid.
-	/// See UIScrollView's [documentation](https://developer.apple.com/documentation/uikit/uiscrollviewaccessibilitydelegate/1621055-accessibilityscrollstatus) for more information.
-	public var scrollViewAccessibilityScrollStatus: ((_ scrollView: UIScrollView) -> String?)?
+	public weak var scrollViewDelegate: UIScrollViewDelegate? {
```
raulriera: Missing documentation 😉

Issue #139, opened 2019-03-25
Show example for custom view in readme


jjjeeerrr111:

jjjeeerrr111:
<img width="376" alt="Screen Shot 2019-03-25 at 4 29 24 PM" src="https://user-images.githubusercontent.com/7354919/54960747-30073200-4f1b-11e9-86c2-bb4b4fc4c5c1.png">

PR #140, opened 2019-03-25
Show example for custom view in readme

jjjeeerrr111:
Resolves https://github.com/Shopify/FunctionalTableData/issues/139

Just wanted to add a specific example for adding custom views in FTD. It was not obvious to me the first time using the sdk

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,9 +138,7 @@ Let's build a view replicating a new iMessage cell
 The first step is to build the custom view:
 
 ```swift
-
 class MessageView: UIView {
-
     let profilePicture = UIImageView()
     let nameLabel = UILabel()
     let messageLabel = UILabel()
@@ -157,49 +155,7 @@ class MessageView: UIView {
     }
 
     func setupConstraints() {
-        profilePicture.translatesAutoresizingMaskIntoConstraints = false
-        addSubview(profilePicture)
-
-        addSubview(nameLabel)
-        addSubview(messageLabel)
-
-        timeStampLabel.translatesAutoresizingMaskIntoConstraints = false
-        addSubview(timeStampLabel)
-
-        let stackView = UIStackView(arrangedSubviews: [nameLabel, messageLabel])
-        stackView.axis = .vertical
-        stackView.alignment = .fill
-        stackView.spacing = 3
-        stackView.distribution = .fill
-        stackView.translatesAutoresizingMaskIntoConstraints = false
-        addSubview(stackView)
-
-        let constraints: [NSLayoutConstraint] = [
-        profilePicture.heightAnchor.constraint(equalToConstant: 60),
-        profilePicture.widthAnchor.constraint(equalToConstant: 60),
-        profilePicture.leadingAnchor.constraint(equalTo: leadingAnchor),
-        profilePicture.topAnchor.constraint(equalTo: topAnchor, constant: 12),
-        profilePicture.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -12),
-        stackView.centerYAnchor.constraint(equalTo: profilePicture.centerYAnchor),
-        stackView.topAnchor.constraint(greaterThanOrEqualTo: topAnchor),
-        stackView.bottomAnchor.constraint(lessThanOrEqualTo: bottomAnchor),
-        stackView.leadingAnchor.constraint(equalTo: profilePicture.trailingAnchor, constant: 12),
-        stackView.trailingAnchor.constraint(lessThanOrEqualTo: timeStampLabel.leadingAnchor, constant: -12),
-        timeStampLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
-        timeStampLabel.topAnchor.constraint(equalTo: topAnchor)
-        ]
-        NSLayoutConstraint.activate(constraints)
-
-        profilePicture.layer.borderWidth = 1
-        profilePicture.layer.borderColor = UIColor.black.cgColor
-        profilePicture.layer.cornerRadius = 30
-        profilePicture.backgroundColor = UIColor.gray
-
-        nameLabel.font = UIFont.systemFont(ofSize: 18, weight: .medium)
-        messageLabel.font = UIFont.systemFont(ofSize: 15, weight: .light)
-
-        timeStampLabel.font = UIFont.systemFont(ofSize: 12, weight: .light)
-        timeStampLabel.setContentHuggingPriority(.required, for: .horizontal)
+        // implement your auto layout code
     }
 }
 ```
@@ -207,7 +163,6 @@ class MessageView: UIView {
 Next, we'll build the `MessageViewState` struct that will take care of setting all the views properties:
 
 ```swift
-
 struct Message {
     let name: String
     let image: UIImage
@@ -216,7 +171,6 @@ struct Message {
 }
 
 struct MessageViewState: Equatable {
-
     let date: Date
     let name: String
     let message: String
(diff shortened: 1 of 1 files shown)


FunctionalTableDataDemo/View Controllers/TableExampleController.swift
```diff
@@ -10,45 +10,44 @@ import UIKit
...
+                        return .deselected
+                }),
+                state: LabelState(text: item),
+                cellUpdater: LabelState.updateView)
+        }
+        
+        functionalData.renderAndDiff([
+            TableSection(key: "section", rows: rows)
+            ])
+    }
```
raulriera: Why? :P

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
 
 At the end of the day `HostCell` is just one of the possible implementations of `CellConfigType`, that's the underlying power of this framework.
 
+### Building custom Cells
+
+In order to build custom cells we will need to have 3 parts:
+
+1. A custom view class
+2. A class to manage the view state
```
raulriera:
```suggestion
2. An object to manage the view state
```

 I don't think any example we have are using classes tho.

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+3. A list builder that defines the `HostCell<View, State, Layout>`
```
raulriera: We don't need this step, this is just the convention we have been using

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+
+Let's build a view replicating a new iMessage cell
+
+<img src="https://user-images.githubusercontent.com/7354919/54960747-30073200-4f1b-11e9-86c2-bb4b4fc4c5c1.png" />
+
+The first step is to build the custom view:
+
+```
```
raulriera:
```suggestion
```swift
```

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+}
+
+static func updateView(_ view: MessageView, state: MessageViewState?) {
+guard let state = state else { return }
+view.timeStampLabel.text = "\(state.date)"
+view.nameLabel.text = state.name
+view.messageLabel.text = state.message
+view.profilePicture.image = state.image
+}
+
```
raulriera: All this code needs formatting. For simplicity (since this is getting pretty long) Why not use directly MessageViewState. instead of wrapping everything into `struct Message`

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+timeStampLabel.topAnchor.constraint(equalTo: topAnchor)
+]
+NSLayoutConstraint.activate(constraints)
+
+profilePicture.layer.borderWidth = 1
+profilePicture.layer.borderColor = UIColor.black.cgColor
+profilePicture.layer.cornerRadius = 30
+profilePicture.backgroundColor = UIColor.gray
+
+nameLabel.text = "FirstName LastName"
```
raulriera: We shouldn't put values here, these are controlled by the State

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+for message in messages {
+// each cell should have a unique key
+rows.append(MessageCell(key: "\(message)-\(UUID().uuidString)-cell", state: MessageViewState(message: message)) { (view, state) in
+
+})
+}
+return TableSection(key: "entire-table", rows: rows)
+}
+}
+```
```
raulriera: I think we can skip the builder, we have an example above right? If not, directly just creating the cell without `MessageListBuilder` would be fine

FunctionalTableDataDemo/View Controllers/TableExampleController.swift
jjjeeerrr111: weird. Let me see if I can remove that

README.md
jjjeeerrr111: I don't see a list builder in the current read me. I think it can be valuable to explain that if you have multiple types of `CellConfigType` it is easy to pass all of that logic to the list builder. What do you think? I can make that clearly in the text

README.md
jjjeeerrr111: I had it with the multiple values at first but thought it would be simpler to explain the relationship between the model and state objects by passing it directly

README.md
jjjeeerrr111: I can put is as optional?

README.md
raulriera: What I meant is that a list builder is not part of FunctionalTableData, is just a way for us to centralize the creating of the cells.

I still don't think we should be coupling Models + ListBuilders into this README. Those two are just tied to how we do things, but may not be how people should do it

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+        profilePicture.layer.borderWidth = 1
+        profilePicture.layer.borderColor = UIColor.black.cgColor
+        profilePicture.layer.cornerRadius = 30
+        profilePicture.backgroundColor = UIColor.gray
+
+        nameLabel.font = UIFont.systemFont(ofSize: 18, weight: .medium)
+        messageLabel.font = UIFont.systemFont(ofSize: 15, weight: .light)
+
+        timeStampLabel.font = UIFont.systemFont(ofSize: 12, weight: .light)
+        timeStampLabel.setContentHuggingPriority(.required, for: .horizontal)
```
raulriera: Maybe we can skip this whole thing and just add a comment?

```
func setupConstraints() {
  // implement your auto layout code
}
```

This function is taking a lot of space and its something unrelated to making Custom Cells

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+
+struct Message {
+    let name: String
+    let image: UIImage
+    let date: Date
+    let text: String
+}
+
+struct MessageViewState: Equatable {
+
```
raulriera:
```suggestion
```

(Same for all the places that there is a line between the declaration and the code) :)

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+}
+```
+
+Finally, the `MessageListBuilder` will accept an array of `Messages` and return the appropriate `TableSection`. We will define a custom `HostCell` called `MessageCell` and pass in our custom `MessageView`, `MessageViewState` and layout
+
+```swift
+
+struct MessageListBuilder {
+
```
raulriera:
```suggestion
```

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+        timeStampLabel.font = UIFont.systemFont(ofSize: 12, weight: .light)
+        timeStampLabel.setContentHuggingPriority(.required, for: .horizontal)
+    }
+}
+```
+
+Next, we'll build the `MessageViewState` struct that will take care of setting all the views properties:
+
+```swift
+
```
raulriera:
```suggestion
```

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+Let's build a view replicating a new iMessage cell
+
+<img src="https://user-images.githubusercontent.com/7354919/54960747-30073200-4f1b-11e9-86c2-bb4b4fc4c5c1.png" />
+
+The first step is to build the custom view:
+
+```swift
+
+class MessageView: UIView {
+
```
raulriera:
```suggestion
```

README.md
```diff
@@ -123,5 +123,146 @@ LabelCell(key: "company", state: LabelState(text: "Shopify",
...
+3. And optionally, a list builder to build the table view sections
+
+Let's build a view replicating a new iMessage cell
+
+<img src="https://user-images.githubusercontent.com/7354919/54960747-30073200-4f1b-11e9-86c2-bb4b4fc4c5c1.png" />
+
+The first step is to build the custom view:
+
+```swift
+
```
raulriera:
```suggestion
```

README.md
jjjeeerrr111: yea thats valid

PR #141, opened 2019-03-26
Updated for Xcode 10.2

krbarnes:
Fixed compile errors introduced by Xcode 10.2

### DIFF
--- a/FunctionalTableData/TableSection.swift
+++ b/FunctionalTableData/TableSection.swift
@@ -20,7 +20,7 @@ public protocol TableSectionType {
 	var rows: [CellConfigType] { get }
 	/// Action to perform when the header view comes in or out of view.
 	var headerVisibilityAction: ((_ view: UIView, _ visible: Bool) -> Void)? { get }
-	@available(iOS, deprecated: 1.0, message: "Use `rows.count` instead.")
+	@available(*, deprecated, message: "Use `rows.count` instead.")
 	var rowCount: Int { get }
 
 	subscript(index: Int) -> CellConfigType { get }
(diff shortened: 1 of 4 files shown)


FunctionalTableData/TableSection.swift
```diff
@@ -20,7 +20,7 @@ public protocol TableSectionType {
 	var rows: [CellConfigType] { get }
 	/// Action to perform when the header view comes in or out of view.
 	var headerVisibilityAction: ((_ view: UIView, _ visible: Bool) -> Void)? { get }
-	@available(*, deprecated: 1.0, message: "Use `rows.count` instead.")
+	@available(iOS, deprecated: 1.0, message: "Use `rows.count` instead.")
```
krbarnes: I don't know why, but Swift now hates "*" and a deprecated version.
Does this actually mean it's been deprecated since iOS 1.0?

dannbeaus: does this work on older Xcode versions? Wondering if we are forcing 10.2 on everyone now.

robinsenior: do we want to constraint FTD to iOS? there's nothing stopping it from working in tvOS is there?

josephroquedev: Yeah, it'll come up as "`function` has been deprecated since iOS 1.0", I've seen this before.

robinsenior: I mean, other than nobody giving a shit about tvOS, that is

josephroquedev: Looks like you could just omit the version number altogether: `@available(iOS, deprecated, message: "message")`

Other options here: https://docs.swift.org/swift-book/ReferenceManual/Attributes.html

krbarnes: There are no tvOS targets right now.

dannbeaus: Builds on 10.1 👍

krbarnes: I had that doc open @josephroquedev. It seems to imply that it was fine the way it's been for the last [value] years. 
I'll remove the version number and put back the *.

* == iOS according the build targets  😉

PR #143, opened 2019-03-29
Support setting the accessoryView as part of CellStyle

abecevello:
`CellStyle` provided access to the `accessoryType` property for UITableViewCells. This change adds the `accessoryView` property to allow custom accessory views.

### DIFF
--- a/FunctionalTableData/CellStyle.swift
+++ b/FunctionalTableData/CellStyle.swift
@@ -50,6 +50,10 @@ public struct CellStyle {
 	///
 	/// Supported by `UITableView` only.
 	public var accessoryType: UITableViewCell.AccessoryType
+	/// The custom view to use for the accessory. If set the accessoryType is ignored.
+	///
+	/// Supported by `UITableView` only.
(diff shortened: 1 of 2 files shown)


g-Off:
I think this will result in every renderAndDiff pass updating every cell.
Each time your Builder runs it'll produce a new UIView for the accessoryView, CellStyles `==` compares them using `==`, which in turn will do pointer equality IIRC, which will result in it thinking that the cell has changed

abecevello:
@g-Off Do you have any suggestions to work around that, or do you think this isn't going to work at all?

josephroquedev:
Could an approach like [`BackgroundViewProvider`](https://github.com/Shopify/FunctionalTableData/blob/master/FunctionalTableData/BackgroundViewProvider.swift) work?

abecevello:
I decided to put the fact I need this image in the state object and make it part of the custom view for the cell. It would be nicer to have it part of the `accessoryView`, but putting it in the custom state and view is the next best option.

This PR is no longer necessary so I'm going to close it.

PR #146, opened 2019-04-03
moves the cell styling logic into its own object

g-Off:
Moves cell styling into its own object type. This consolidates logic that was happening in a few locations, also keeps track of a cell being highlighted.

### DIFF
--- a/FunctionalTableData/Extensions/Array+TableSection.swift
+++ b/FunctionalTableData/Extensions/Array+TableSection.swift
@@ -28,9 +28,24 @@ extension Array where Element: TableSectionType {
 	}
 }
 
+extension Array where Element: TableSectionType {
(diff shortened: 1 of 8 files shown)


FunctionalTableData/Extensions/Array+TableSection.swift
```diff
@@ -28,9 +28,24 @@ extension Array where Element: TableSectionType {
 	}
 }
 
+extension Array where Element: TableSectionType {
+	func indexPath(from keyPath: FunctionalTableData.KeyPath) -> IndexPath? {
+		if let sectionIndex = self.index(where: { $0.key == keyPath.sectionKey }), let rowIndex = self[sectionIndex].rows.index(where: { $0.key == keyPath.rowKey }) {
```
raulriera: Hit the migrate to Swift 5 button :D

PR #147, opened 2019-04-03
Simple Cell Examples

raulriera:
## What

In order to make FunctionalTableData easier to use, introducing a few example cells to get up and running very fast. The current cells are now provided in the demo project:

- Label
- Button
- Image
- Spacer
- Subtitle (title and subtitle labels vertically)
- Switch
- Combined (allows to use a left and side cell of any combination)

Also includes a screen resembling the Settings screen from iOS, worried people would think this library is only for creating "form" layouts tho 🤔 

Screen Shot 2019-04-03 at 1 56 56 PM


FunctionalTableDataDemo/Cells/UIControl+Extensions.swift
```diff
@@ -0,0 +1,83 @@
+//
+//  UIControl+Extensions.swift
+//  Shopify
+//
+//  Created by Tom Burns on 2016-01-06.
```
raulriera: @boourns

raulriera:
I'm thinking of removing the tableview and collection view examples

FunctionalTableDataDemo/Cells/UIControl+Extensions.swift
boourns: I figured you would have deleted all my code by now! :D

FunctionalTableDataDemo/AppDelegate.swift
```diff
@@ -7,13 +7,19 @@
...
+import FunctionalTableData
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
 	var window: UIWindow?
 
 	func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+		
+		Separator.appearance().backgroundColor = UITableView().separatorColor
```
g-Off: initializing a tableview to get the separate colour? 🤢

FunctionalTableDataDemo/Cells/Button/ButtonState.swift
```diff
@@ -0,0 +1,45 @@
...
+import UIKit
+import FunctionalTableData
+
+typealias ButtonCell = HostCell<UIButton, ButtonState, LayoutMarginsTableItemLayout>
+
+public struct ButtonState: Equatable {
+	public let title: String
+	public let isEnabled: Bool
+	public let alignment: UIControl.ContentHorizontalAlignment
+	public let action: (UIButton) -> Void
```
g-Off: Might be able to get away with a target/action approach instead of closure based.
A `nil` target with an action that exists on the controller that the view will be inside should, I think, fire correctly (and TBH is a nice demo of how the responder chain works/was intended to work).
Alternatively an actual target/action object/selector captured here could also be nicer (target held onto weakly, state comparison with == can then actually compare the target via `===` (identity) and the action directly).

Haven't tried either of these approaches, but my opinion is that they lead to better leveraging of the Cocoa APIs, and they also lead in a direction where the state instances can be made Codable (minus the target pointing to an instance, that's not going to work)

g-Off: Could also ignore this comment for now and try it after merging...

FunctionalTableDataDemo/AppDelegate.swift
raulriera: Is the only way to get the real default value 😔 but I will use a random Gray color instead so this idea doesn't spread

FunctionalTableDataDemo/Cells/Button/ButtonState.swift
raulriera: Would you delete the backing UIControl extension as well? That will completely remove all the code Tom built 😏

PR #157, opened 2019-04-10
Add separator style inset on both sides

abecevello:
We have a new design for separators that are inset on both sides (so they appear in the centre). This adds a new style to `Separator` called `bothInset` (naming suggestions very welcome!) that anchors the separator to the layout margins on both sides.

Technically adding a new enum value is a breaking change since it would require anyone that uses it in a switch statement without a `default` to add the new case, but I feel like most people aren't checking this enum in a switch statement outside of FTD.

### DIFF
--- a/FunctionalTableData/Separator.swift
+++ b/FunctionalTableData/Separator.swift
@@ -21,20 +21,25 @@ public class Separator: UIView {
 	
 	/// The style for table cells used as separators.
 	///
-	/// The options are `full`, `inset`, and `moreInset`
-	public enum Style {
+	/// The options are `full`, `inset`, `moreInset`, `custom`
+	public enum Style: Equatable {
 		case full
 		case inset
 		case moreInset
+		case custom(leadingInset: CGFloat, trailingInset: CGFloat, layoutMarginsRelative: Bool)
 		
-		public var insetDistance: CGFloat {
-			switch self {
-			case .inset:
-				return Separator.inset
-			case .full:
-				return 0
-			case .moreInset:
-				return 3 * Separator.inset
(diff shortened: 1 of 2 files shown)


abecevello:
Hold off on reviewing this. I'm going to need to make this more configurable.

abecevello:
This has been updated to allow specifying a custom separator inset. This will allow for more configurability and allow for the full range of new designs we have.

FunctionalTableData/Separator.swift
```diff
@@ -21,11 +21,12 @@ public class Separator: UIView {
...
 	/// The style for table cells used as separators.
 	///
-	/// The options are `full`, `inset`, and `moreInset`
-	public enum Style {
+	/// The options are `full`, `inset`, `moreInset`, `custom`
+	public enum Style: Equatable {
 		case full
 		case inset
 		case moreInset
+		case custom(leftInset: CGFloat, rightInset: CGFloat, layoutMarginsRelative: Bool)
```
Caiopia: nit: maybe could name these `leadingInset` and `trailingInset` to match the constraints naming below?

```suggestion
		case custom(leadingInset: CGFloat, trailingInset: CGFloat, layoutMarginsRelative: Bool)
```

FunctionalTableData/Separator.swift
```diff
@@ -21,11 +21,12 @@ public class Separator: UIView {
...
 		
 		public var insetDistance: CGFloat {
```
g-Off: I think we could get rid of this public var instead. It's only used in one spot and I was a bit confused about the .custom being 0 when .custom actually has values associated with it

abecevello: Removed.

PR #159, opened 2019-04-16
Adds the isSameKind function

g-Off:
Moves the `type(of:)` comparisons into an overridable function that can be used in tests.
Captures the type of a cell in its `debugInfo()`

With this change I'm able to simulate different "types" of cells in test cases by having a custom type that conforms to CellConfigType and has the `isSameKind` function implemented. Additionally, capturing the type of a cell in a crash means that that data can be used in conjunction with the isSameKind

### DIFF
--- a/FunctionalTableData/CellConfigType.swift
+++ b/FunctionalTableData/CellConfigType.swift
@@ -45,4 +45,19 @@ public protocol CellConfigType: TableItemConfigType, CollectionItemConfigType {
 	/// - Returns: `true` when both values are the same, `false` otherwise.
 	func isEqual(_ other: CellConfigType) -> Bool
 	func debugInfo() -> [String: Any]
+	
+	/// Compares self against another `CellConfigType` to determine if they should be counted as being the same type.
+	///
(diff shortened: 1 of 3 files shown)


FunctionalTableData/CellConfigType.swift
```diff
@@ -45,4 +45,19 @@ public protocol CellConfigType: TableItemConfigType, CollectionItemConfigType {
 	/// - Returns: `true` when both values are the same, `false` otherwise.
 	func isEqual(_ other: CellConfigType) -> Bool
 	func debugInfo() -> [String: Any]
+	
+	/// Compares self against another `CellConfigType` to determine if they should be counted as being the same type.
+	///
+	/// - note: This should almost never need to be implemented. The default version of it, provided as a protocol extension,
+	///   does the expected thing (and compares `type(of:)` to `type(of:)`. This is inteded for test cases to be able to override.
```
g-Off:
```suggestion
	///   does the expected thing (and compares `type(of:)` to `type(of:)`. This is intended for test cases to be able to override.
```

PR #160, opened 2019-04-17
Fix invalid index paths on reload and an incorrect compare for updates

g-Off:
- when comparing the old/new types, compare against the proper old index path
- reloads require the before IndexPath, fixes to use that instead of the new IndexPath


raulriera:
Tests didn't break? :fry-eyes:

PR #161, opened 2019-04-18
Adds the ability to integrate with collection/table view prefetch

g-Off:
Prefetch operations are requested from CellActions and executed in a private OperationQueue. The actual fetching/cancelling is delayed by a short period of time because iOS may trigger rapid fetch/cancel operations and this is an attempt to reduce the amount of fetch/cancel/fetch of the same IndexPath

This PR introduces a common prefetch data source that's usable by both collection and tableview's. They essentially have the same delegate methods, just passing in a UICollectionView in one and a UITableView in the other. Object conforms to both, takes advantage of the same mechanism for queueing and dequeuing operations/cancelling/etc.


FunctionalTableData/DataSourcePrefetching.swift
```diff
@@ -0,0 +1,152 @@
...
+		operationItemPathsToPrefetch.removeAll()
+		operationItemPathsToCancel.removeAll()
+		
+		queue.addOperations(newOperations, waitUntilFinished: false)
+		cancelledOperations.forEach { $0.operation.cancel() }
+	}
+}
+
+private extension NSLocking {
+	func withLock<T> (_ body: () throws -> T) rethrows -> T {
```
raulriera: I was wondering about this `withLock`. I t thought it was a new thing in iOS

FunctionalTableDataTests/DataPrefetchTests.swift
```diff
@@ -0,0 +1,97 @@
...
+		
+		override func cancel() {
+			super.cancel()
+			cancelExpectation?.fulfill()
+		}
+	}
+	
+	private func section(withPrefetch prefetchAction: @escaping CellActions.PrefetchAction) -> TableSection {
+		let actions = CellActions(prefetchAction: prefetchAction)
+		let cell = HostCell<UIView, String, LayoutMarginsTableItemLayout>(key: "prefetch-cell", actions: actions, state: "Actions") { (_, _) in }
```
raulriera: Pretty sure we have some `TestCell` or something out there

PR #163, opened 2019-04-24
Move sections to a TableData class instead of duplicating the array

PierreOleo:
In the current implementation, when the user manually moves a row `tableView(_ tableView: UITableView, moveRowAt sourceIndexPath: IndexPath, to destinationIndexPath: IndexPath)` will modify the datasource `sections` but not the `sections` of `FunctionalTableData`. 

If the TableSection `didMoveRow` calls for a re-render then `FunctionalTableData` will detect changes. If the changes detected are "move" changes in an animated render, another unexpected move could be applied on top of the current one

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData.swift
@@ -41,20 +41,18 @@ public class FunctionalTableData {
 	
 	private func dumpDebugInfoForChanges(_ changes: TableSectionChangeSet, previousSections: [TableSection], visibleIndexPaths: [IndexPath], exceptionReason: String?, exceptionUserInfo: [AnyHashable: Any]?) {
 		guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
-		let exception = Exception(name: name, newSections: sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: tableView?.frame ?? .zero, reason: exceptionReason, userInfo: exceptionUserInfo)
+		let exception = Exception(name: name, newSections: data.sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: tableView?.frame ?? .zero, reason: exceptionReason, userInfo: exceptionUserInfo)
 		exceptionHandler.handle(exception: exception)
 	}
 	
-	var sections: [TableSection] {
-		get {
-			return dataSource.sections
-		}
-		set {
-			cellStyler.sections = newValue
-			dataSource.sections = newValue
-			delegate.sections = newValue
+	private var data: TableData {
+		didSet {
+			cellStyler.data = data
+			dataSource.data = data
+			delegate.data = data
 		}
 	}
+
 	private static let reloadEntireTableThreshold = 20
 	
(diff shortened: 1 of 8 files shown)


FunctionalTableDataDemo/View Controllers/TableExampleController.swift
```diff
@@ -11,28 +11,25 @@ import FunctionalTableData
...
-		items.append(NSDate().description)
+		items.append("\(items.count)")
+		render()
 	}
 	
 	private func render() {
 		let rows: [CellConfigType] = items.enumerated().map { index, item in
 			return LabelCell(
-				key: "id-\(index)",
+				key: "id-\(item)",
```
dannbeaus: why this change? index helps with uniquing, I don't suggest making that change.

PierreOleo: Will remove this change

The intention was to have a manual sorting simple sample code with moves detection. Keeping the index as an id makes keys vary for each object on each render when reordering. FunctionalTableData then detects an insertion and a deletion instead of a move.
Uniquing was possible as there is no way to delete a item from the array.

But not sure of the added value anymore, moreover with github.com/Shopify/FunctionalTableData/pull/147  removing this class really soon.

raulriera: With the item it will crash if we spam the button right? But I agree with not using index here. I don't think it matters. I want to delete these examples anyway

raulriera:
You had a previous implementation, what happened?

PierreOleo:
@g-Off suggested the `TableData` approach as it would be cleaner than just passing arrays everywhere

FunctionalTableDataDemo/View Controllers/TableExampleController.swift
PierreOleo: Should not have crashed as `NSDate().description` was replaced by `items.count` but decided to revert the example changes

FunctionalTableData/TableView/FunctionalTableData.swift
```diff
@@ -41,17 +41,18 @@ public class FunctionalTableData {
 	
 	private func dumpDebugInfoForChanges(_ changes: TableSectionChangeSet, previousSections: [TableSection], visibleIndexPaths: [IndexPath], exceptionReason: String?, exceptionUserInfo: [AnyHashable: Any]?) {
 		guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
-		let exception = Exception(name: name, newSections: sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: tableView?.frame ?? .zero, reason: exceptionReason, userInfo: exceptionUserInfo)
+		let exception = Exception(name: name, newSections: data.sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: tableView?.frame ?? .zero, reason: exceptionReason, userInfo: exceptionUserInfo)
 		exceptionHandler.handle(exception: exception)
 	}
 	
-	private var sections: [TableSection] = [] {
+	private var data: TableData {
```
g-Off: This can be a let now can't it? Should also be able to get rid of the didSet portion completely as well

PR #170, opened 2019-05-22
Separator customization

raulriera:
## What

This PR allows for more control over the separator in FunctionalTableData. This will be a breaking change for anyone using `.custom`. But it's safe for `.inset` and `.full`.

The demo project shouldn't visually change, and the tests have been modified to account for the new styling.

### DIFF
--- a/FunctionalTableData/Separator.swift
+++ b/FunctionalTableData/Separator.swift
@@ -14,33 +14,37 @@ import UIKit
 ///
 /// Supported by `UITableView` only.
 public class Separator: UIView {
-	/// Specifies the default inset of cell separators.
-	public static var inset: CGFloat = 0.0
-	/// Specifies the default thickness of cell separators.
-	private let thickness: CGFloat = 1.0 / UIScreen.main.scale
-	
 	/// The style for table cells used as separators.
-	///
-	/// The options are `full`, `inset`, `moreInset`, `custom`
-	public enum Style: Equatable {
-		case full
-		case inset
-		case moreInset
-		case custom(leadingInset: CGFloat, trailingInset: CGFloat, layoutMarginsRelative: Bool)
-		
-		public static func ==(lhs: Style, rhs: Style) -> Bool {
-			switch (lhs, rhs) {
-			case let (.custom(lhsLeadingInset, lhsTrailingInset, lhsLayoutMargins), .custom(rhsLeadingInset, rhsTrailingInset, rhsLayoutMargins)):
-				return lhsLeadingInset == rhsLeadingInset && lhsTrailingInset == rhsTrailingInset && lhsLayoutMargins == rhsLayoutMargins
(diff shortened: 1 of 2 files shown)


FunctionalTableData/Separator.swift
```diff
@@ -14,34 +14,38 @@ import UIKit
...
+		
+		/// Specifies the leading inset of the separators.
+		public let leadingInset: Inset
+		/// Specifies the trailing inset of the separators.
+		public let trailingInset: Inset
+		/// Specifies the thickness of cell separators.
+		public let thickness: CGFloat
+		
+		static public let full: Style = Style(leadingInset: .none, trailingInset: .none)
+		static public let inset: Style = Style(leadingInset: .init(value: 0, respectingLayoutMargins: true), trailingInset: .none)
```
raulriera: These two values allows for a pretty nice usage of the separator. But making custom ones is a little cumbersome. Open to suggestions. 

```swift
CellStyle(bottomSeparator: Separator.Style(leadingInset: .init(value: 10, respectingLayoutMargins: false), trailingInset: .none, thickness: 20))
```

oof 😅

FunctionalTableData/Separator.swift
```diff
@@ -14,34 +14,38 @@ import UIKit
...
+		
+		public init(leadingInset: Inset, trailingInset: Inset, thickness: CGFloat = 1.0) {
+			self.leadingInset = leadingInset
+			self.trailingInset = trailingInset
+			self.thickness = thickness / UIScreen.main.scale
```
raulriera: Too magic? people won't know this is happening. Tripped me for a second in the tests

FunctionalTableData/Separator.swift
raulriera: This? I not a fan of enums 😄 (for no real reason)

```swift
enum Insets {
	case edge(CGFloat)
	case margins(CGFloat)
}
```

Caiopia: I think the enum would shorten it the most, but we could also put default values in the `Style` and `Inset` init methods that could make it less verbose for some cases.

raulriera: which defaults values?

```swift
static func edge(value: CGFloat) -> Style {}
```

Something like that?

Caiopia: No I was thinking having a default for `respectingLayoutMargins` in the init method, but that only shortens it for 1 case 😜. I think enum could improve things...

```swift
CellStyle(bottomSeparator: Separator.Style(leadingInset: .edge(10), trailingInset: .margins(10), thickness: 20))
```

raulriera: The static func would work the exact same way. Apple already does similar things with CALayer and Shapes. I will circle that later when people complain 😄

PR #172, opened 2019-06-24
Allow cells to be unhighlighted

krbarnes:
When using CanSelectAction you should be able to cancel the selection and have the highlight cell revert to the previous one or none if nothing was selected. The code currently highlights the cell even if you cancel the selection. 

This PR saves the original highlight cell and sets it to be highlighted again if the selection is cancelled.

### DIFF
--- a/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
+++ b/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
@@ -93,6 +93,7 @@ extension FunctionalTableData {
 				return nil
 			}
 			
+			let highlightRow = cellStyler.highlightedRow
 			let keyPath = data.sections.itemPath(from: indexPath)
 			
 			if let canSelectAction = cellConfig.actions.canSelectAction {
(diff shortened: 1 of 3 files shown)


FunctionalTableDataDemo/View Controllers/TableExampleController.swift
```diff
@@ -99,5 +100,41 @@ class TableExampleController: UITableViewController {
...
+			callback(false)
+			self.dismiss(animated: true)
+		})
+		alertController.addAction(UIAlertAction(title: "OK", style: .default) { [weak self] alert in
+			guard let self = self else { return }
+			callback(true)
+			self.dismiss(animated: true)
+		})
+		present(alertController, animated: true)
+	}
```
krbarnes: This is my test code.. Shouldn't merged as is because it doesn't look good.

FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
```diff
@@ -105,7 +106,7 @@ extension FunctionalTableData {
 						self.tableView(tableView, didSelectRowAt: indexPath)
 						NotificationCenter.default.post(name: UITableView.selectionDidChangeNotification, object: tableView)
 					} else {
-						self.cellStyler.highlightRow(at: keyPath, animated: false, in: tableView)
+						self.cellStyler.highlightRow(at: previousHighlightPath, animated: true, in: tableView)
```
raulriera: Why flip the animated?

krbarnes: Fake news. Never happened.

FunctionalTableDataDemo/View Controllers/TableExampleController.swift
krbarnes: improved code.. comment no longer valid

PR #174, opened 2019-06-28
Fix bug where the Section Headers are not updated correctly

mmroz:
Fixed equality checking for the TableSectionHeaderFooter so that it changes appropriately.

### DIFF
--- a/FunctionalTableData/TableSectionHeaderFooter.swift
+++ b/FunctionalTableData/TableSectionHeaderFooter.swift
@@ -19,13 +19,7 @@ public protocol TableHeaderFooterConfigType: TableItemConfigType {
 	var height: CGFloat { get }
 }
 
-public extension TableHeaderFooterConfigType {
-	func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool {
-		return (other as? Self) != nil
-	}
-}
-
-public protocol TableHeaderFooterStateType {
+public protocol TableHeaderFooterStateType: Equatable {
 	var insets: UIEdgeInsets { get }
 	var height: CGFloat { get }
 	var topSeparatorHidden: Bool { get }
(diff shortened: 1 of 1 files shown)


Campbell-Scott:
Some context on why this is here:

We wanted an `isEqual(_ other: TableHeaderFooterConfigType?) -> Bool` so we could compare two header/footers and re-render a `TableSection` if there were any state changes. If we added this as a requirement to the `TableHeaderFooterConfigType` protocol, without providing a default implementation, it would become a breaking change for any objects conforming to `TableHeaderFooterConfigType` that didn't implement it. The default implementation only exists to prevent that (it should always just evaluate too `true`, if the headers are of the same type, so that we maintain the previous behaviour of not updating at all).

This fix implements the functionality for the out-of-the-box `TableSectionHeaderFooter` that we provide in ftd, but it removes the default conformance for any other custom implementors of `TableHeaderFooterConfigType`; which is a breaking change.

FunctionalTableData/TableSectionHeaderFooter.swift
```diff
@@ -19,13 +19,7 @@ public protocol TableHeaderFooterConfigType: TableItemConfigType {
 	var height: CGFloat { get }
 }
 
-public extension TableHeaderFooterConfigType {
-	func isEqual(_ other: TableHeaderFooterConfigType?) -> Bool {
-		return (other as? Self) != nil
-	}
-}
```
Caiopia: Removing this is potentially a breaking change for protocol adopters who relied on the default implementation. It should be enough to just override the method like you did below, while keeping this one here as the default.

g-Off: Breaking changes are fine, master is currently in a state of broken changes as it works towards its 2.0 release. If you need non-breaking changes then you stick to the tagged release

Caiopia: Oh yeah, that's right. Though I still think we should keep the default implementation.

g-Off: > Oh yeah, that's right. Though I still think we should keep the default implementation.

I... disagree. The default one isn't great and we don't have a default for CellConfigType's isEqual

Caiopia: Fair enough, no strong opinion.

PR #176, opened 2019-07-24
Add delaysContentTouches to CellStyle

stevenwenxu:
When a table view row has a button in it, the button won't have an immediate highlighted tapping state because the system needs to determine if the tapping was part of a scroll. The tapping state only appears when the button is long-pressed for 1 second or so. 

This PR adds a configurable `delaysContentTouches`, so that users of FTD have control over whether they want this behaviour or not. When this parameter is not used, it has the default system behaviour.

It appears that starting from iOS 7, each UITableViewCell has its own internal scroll view which interferes with the flag set on the table view, and it's been suggested by [the community](https://stackoverflow.com/a/22925576) to do this on the cell level as well for the desired behaviour.


stevenwenxu:
It looks like this isn't needed for now.

PR #177, opened 2019-08-22
Improve diff performances with an early return for empty cases

TheoBdev:
By avoiding to run the diff algorithm in full in case of old and/or new sections empty, we gained a few milliseconds in the ChangeSet `calculateChanges` function.

The performances tests have also been moved to a separate file.


FunctionalTableDataTests/FunctionalDataTests.swift
```diff
@@ -90,27 +90,6 @@ class FunctionalDataTests: XCTestCase {
 		let cell = tableData.tableView?.visibleCells.first
 		XCTAssertEqual(cell?.accessibilityIdentifier, "sectionKey.cellKey")
 	}
-	
-	func testPerformance() {
```
TheoBdev: This became `testPerformanceByKeepingRows`

PR #178, opened 2019-09-06
Add proper equality for TableSection

abecevello:
Currently TableSection equality conformance only checks if the keys are equal. Although this specific equality function isn't used during FTD diffing, the fact this equality function is incomplete is causing sections to fail to update in some limited cases.

This PR implements proper equality of TableSections by checking all properties, including all rows.

### DIFF
--- a/FunctionalTableData/TableSection.swift
+++ b/FunctionalTableData/TableSection.swift
@@ -164,6 +164,20 @@ extension Array where Element: TableSectionType {
 
 extension TableSection: Equatable {
 	public static func ==(lhs: TableSection, rhs: TableSection) -> Bool {
-		return lhs.key == rhs.key
+		return lhs.key == rhs.key &&
+			(lhs.header == nil && rhs.header == nil || lhs.header?.isEqual(rhs.header) ?? false) &&
+			(lhs.footer == nil && rhs.footer == nil || lhs.footer?.isEqual(rhs.footer) ?? false) &&
+			isEqual(lhs: lhs.rows, rhs: rhs.rows) &&
+			lhs.style == rhs.style &&
+			(lhs.headerVisibilityAction == nil && rhs.headerVisibilityAction == nil || lhs.headerVisibilityAction != nil && rhs.headerVisibilityAction != nil) &&
+			(lhs.didMoveRow == nil && rhs.didMoveRow == nil || lhs.didMoveRow != nil && rhs.didMoveRow != nil)
+	}
+}
+
+private func isEqual(lhs: [CellConfigType], rhs: [CellConfigType]) -> Bool {
+	guard lhs.count == rhs.count else { return false }
+	
+	return zip(lhs, rhs).allSatisfy { (leftCell, rightCell) -> Bool in
+		return leftCell.isEqual(rightCell)
 	}
 }
--- a/FunctionalTableData.xcodeproj/project.pbxproj
+++ b/FunctionalTableData.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		02F106002326B9DF009865B0 /* TableSectionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02F105FF2326B9DF009865B0 /* TableSectionTests.swift */; };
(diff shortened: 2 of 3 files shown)


FunctionalTableData/TableSection.swift
```diff
@@ -164,6 +164,20 @@ extension Array where Element: TableSectionType {
 
 extension TableSection: Equatable {
 	public static func ==(lhs: TableSection, rhs: TableSection) -> Bool {
-		return lhs.key == rhs.key
+		return lhs.key == rhs.key &&
+			(lhs.header == nil && rhs.header == nil || lhs.header?.isEqual(rhs.header) ?? false) &&
```
raulriera: Why are you checking for the nil? Wouldn't isEqual do this too?

FunctionalTableData/TableSection.swift
```diff
@@ -164,6 +164,20 @@ extension Array where Element: TableSectionType {
...
+			(lhs.footer == nil && rhs.footer == nil || lhs.footer?.isEqual(rhs.footer) ?? false) &&
+			isEqual(lhs: lhs.rows, rhs: rhs.rows) &&
```
raulriera: The equality of the rows happen somewhere else, no need to do this here right?

FunctionalTableData/TableSection.swift
```diff
@@ -164,6 +164,20 @@ extension Array where Element: TableSectionType {
...
+			lhs.style == rhs.style &&
+			(lhs.headerVisibilityAction == nil && rhs.headerVisibilityAction == nil || lhs.headerVisibilityAction != nil && rhs.headerVisibilityAction != nil) &&
+			(lhs.didMoveRow == nil && rhs.didMoveRow == nil || lhs.didMoveRow != nil && rhs.didMoveRow != nil)
```
raulriera: You shouldn't include the closures in the equality checks

FunctionalTableData/TableSection.swift
abecevello: `TableHeaderFooterConfigType` is not equatable but has a `isEqual` method defined for it. This means the `isEqual` method call has to handle nil here.

FunctionalTableData/TableSection.swift
abecevello: The `==` function is only used in 2 places in a closed source framework built on this framework. In order to ensure this equality function as defined remains correct, it needs to do row equality checks here. You are correct that in normal operation the diffing algorithm handles row comparisons separately outside of this equality function, but I think that might just be due to differing times that the diffing and this function were created.

FunctionalTableData/TableSection.swift
raulriera: What I mean is why not just use this

`lhs.header?.isEqual(rhs.header)`. Wouldn't isEqual return false if one of them is `nil`?

FunctionalTableData/TableSection.swift
abecevello: @g-Off said to include them like this, but I can remove them.

FunctionalTableData/TableSection.swift
raulriera: Wouldn't this essentially run them twice? Equality check on all rows when checking the TableSection equality and then equality check on all rows in `TableSectionChangeSet`?

FunctionalTableData/TableSection.swift
abecevello: If one of them is nil and the other isn't then `isEqual` should return false, but if both are nil then `isEqual` should return true. That line of code you suggest doesn't handle that case.

FunctionalTableData/TableSection.swift
abecevello: I ran the code and this `==` function isn't run during the diffing algorithm. I commented out this entire function and it was only used in 2 places, none of which were in the FunctionalTableData framework.

raulriera: 😳

FunctionalTableData/TableSection.swift
g-Off: A nil `didMoveRow` to non-nil, or vice-versa, should could as not being equal (same for the `headerVisibilityAction`)

g-Off:
Can we add a couple of unit tests as well?

FunctionalTableData/TableSection.swift
abecevello: Added back.

abecevello:
@g-off Tests have been added

PR #179, opened 2019-09-12
Add duplicated keys to failing log

YasMarcu:
This makes life easier to debug.

### DIFF
--- a/FunctionalTableData/Extensions/Array+TableSection.swift
+++ b/FunctionalTableData/Extensions/Array+TableSection.swift
@@ -11,7 +11,7 @@ extension Array where Element: TableSectionType {
 		let sectionKeys = map { $0.key }
 		if Set(sectionKeys).count != count {
 			let dupKeys = duplicateKeys()
-			let reason = "\(senderName) : Duplicate Section keys"
+			let reason = "\(senderName) : Duplicate Section keys \(dupKeys)"
 			let userInfo: [String: Any] = ["Duplicates": dupKeys]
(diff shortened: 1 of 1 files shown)


raulriera:
The keys are in the `userInfo`, any reason why it doesn't work for you there?

YasMarcu:
@raulriera So the `userInfo` is not printed in the console, only the `reason` is. userInfo is used as a dict data source for the receiver.

As you can see here, only the reason was printed and not the userInfo:
```
2019-09-13 11:40:33.969657-0400 POSNext[6234:2156601] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'DeviceDetailsController : Duplicate Section keys'
```

g-Off:
This leads to undesirable behaviour in crash reporting as no two crashes are considered the same anymore because the crash reason will vary. We used to have it like this but this resulted in our crash reporter having hundreds of unique reports instead of a single one with varying data.

Is it during debugging that you're trying to output more information to the console about the exception reason?

YasMarcu:
@g-Off I'm trying to output more info for both `Debug` and `Release` mode. 
1) `Debug`: Of course for debug mode, there's always the alternative, which is printing the dupKeys in console in case of a crash. Extra effort but no biggy.
2) `Release`: regarding release mode and your crash report concern, why don't we use the NSException name property as the uniquely defined exception. Also my concern is in the crash report, isn't useful to see both the senderName and duplicated key? 
```
let reason = "\(senderName) : Duplicate Section keys \(dupKeys)"
```

I'm just trying to understand 😅

YasMarcu:
@krbarnes I see! That's annoying. Thanks for the explanation. I'll close the PR.

PR #180, opened 2019-09-15
Support for SPM

raulriera:
# Description

Adds support for Swift Package Manager. Also fixed some compile errors Xcode 11 will find in this project (when trying to compile it on its own)

~~Marking this as WIP until I solve the lack of Objective-C bridging header~~ Solved with a dependency that was doing pretty much the same thing as us.

I plan to make a new repo for the example using this package

~~CI support is not quite yet in GitHub maybe? https://github.com/Shopify/FunctionalTableData/pull/180/checks?check_run_id=245792419#step:3:4~~


Package.swift
```diff
@@ -0,0 +1,33 @@
...
+            dependencies: [
+				"CwlCatchException"
+		]),
+        .testTarget(
+            name: "FunctionalTableDataTests",
+            dependencies: [
+				"FunctionalTableData"
+			]),
+    ]
+)
```
raulriera: Indentation 👨‍🍳💋

Sources/FunctionalTableData/CellActions.swift
```diff
@@ -72,7 +72,7 @@ public struct CellActions {
 				return rowAction
 			}
 			
-			@available(iOSApplicationExtension 11.0, *)
```
dannbeaus: Wonder why it was this way.

raulriera: Unsure as well, maybe a blanket replace to match our other frameworks. But Xcode 11 wasn’t having ANY of that now 😂

Package.resolved
```diff
@@ -0,0 +1,16 @@
+{
+  "object": {
+    "pins": [
+      {
+        "package": "CwlCatchException",
+        "repositoryURL": "https://github.com/mattgallagher/CwlCatchException.git",
+        "state": {
+          "branch": null,
+          "revision": "f809deb30dc5c9d9b78c872e553261a61177721a",
+          "version": "2.0.0-beta.1"
```
krbarnes: Any reason we're targeting a beta instead of stable version?

raulriera: Version 2 is what adds support for SPM, there is no stable version of 2 yet

Sources/FunctionalTableData/CellStyle.swift
```diff
@@ -138,7 +138,7 @@ public struct CellStyle {
 		
 		cell.tintColor = tintColor
 		
-		if #available(iOSApplicationExtension 11.0, *) {
+		if #available(iOS 11.0, *) {
```
Caiopia: How come we need this?

dannbeau:
> I plan to make a new repo for the example using this package

was this ever done? if so, what is this repo?

raulriera:
> was this ever done? if so, what is this repo?

No, it wasn't. I'll make one

dannbeau:
Let me know if we can help with it.

PR #184, opened 2019-11-05
Allow configuring masksToBounds in CellStyle

abecevello:
Updates CellStyle to allow configuring the `masksToBounds` property of a cell's layer. This is currently hardcoded to `masksToBounds=true` for UICollectionViewCells. This prevents some animations in cells that can occasionally go slightly outside the cell's bounds.

The default value for this property has been maintained to the existing behaviour.

### DIFF
--- a/Sources/FunctionalTableData/CellStyle.swift
+++ b/Sources/FunctionalTableData/CellStyle.swift
@@ -65,6 +65,10 @@ public struct CellStyle {
 	public var layoutMargins: UIEdgeInsets?
 	/// The radius to use when drawing rounded corners in the view.
 	public var cornerRadius: CGFloat
+	/// Whether the cell's layer is masked to bounds.
+	///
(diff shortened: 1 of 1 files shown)


Sources/FunctionalTableData/CellStyle.swift
```diff
@@ -77,7 +81,8 @@ public struct CellStyle {
 	            backgroundView: UIView?,
 	            tintColor: UIColor? = nil,
 	            layoutMargins: UIEdgeInsets? = nil,
-	            cornerRadius: CGFloat = 0) {
+	            cornerRadius: CGFloat = 0,
+				masksToBounds: Bool = true) {
```
Caiopia:
```suggestion
	            masksToBounds: Bool = true) {
```

raulriera: Did a power move here

PR #185, opened 2019-11-05
bringing back the Xcode project for the time being with inline copy of the exception catching/rethrowing

g-Off:


Sources/FunctionalTableData/ObjCExceptionRethrower.swift
```diff
@@ -7,7 +7,19 @@
 //
 
 import Foundation
+#if SWIFT_PACKAGE
 import CwlCatchException
+#else
```
raulriera: TIL

g-Off:
> We could also just revert my SPM changes

Nah, would rather keep moving forward with that, this is the bandage for now

PR #186, opened 2019-11-05
Fix subscript crash

Kiesco08:
This changes makes sure that the index path is not empty when trying to access `.section`, which would result in a crash otherwise.

### DIFF
--- a/Sources/FunctionalTableData/TableSection.swift
+++ b/Sources/FunctionalTableData/TableSection.swift
@@ -155,7 +155,20 @@ extension TableSection {
 
 extension Array where Element: TableSectionType {
 	subscript(key: IndexPath) -> CellConfigType? {
-		guard key.section < count else { return nil }
+		/// Accessing `.section` or `.row` from an `IndexPath` causes a crash
+		/// if it doesn't have exactly two elements.
+		///
+		/// This can occur with the following steps:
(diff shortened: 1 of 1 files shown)


Sources/FunctionalTableData/TableSection.swift
```diff
@@ -155,7 +155,9 @@ extension TableSection {
 
 extension Array where Element: TableSectionType {
 	subscript(key: IndexPath) -> CellConfigType? {
-		guard key.section < count else { return nil }
+		/// Accessing `.section` or `.row` from an `IndexPath` causes a crash
+		/// if it doesn't have exactly two elements.
+		guard key.count == 2, key.section < count else { return nil }
```
Caiopia: Just found the [precondition of  `count == 2`](https://github.com/apple/swift/blob/6e7051eb1e38e743a514555d09256d12d3fec750/stdlib/public/Darwin/UIKit/UIKit_FoundationExtensions.swift.gyb#L33) that would cause a crash.

Kiesco08: A [feedback](https://feedbackassistant.apple.com/feedback/7431299) was submitted to Apple with the issue, but it seems like I'm the only one who has access to it.

PR #190, opened 2019-12-24
Moved was using the wrong equatable

raulriera:
Small fix that has been lurking around for a while. Thanks to Toussaint Stewart for catching this!

### DIFF
--- a/Sources/FunctionalTableData/TableSectionChangeSet.swift
+++ b/Sources/FunctionalTableData/TableSectionChangeSet.swift
@@ -11,10 +11,6 @@ import Foundation
 struct Moved<T: Equatable>: Equatable {
 	let from: T
 	let to: T
-
-	static func ==(lhs: Moved<T>, rhs: Moved<T>) -> Bool {
-		return lhs.from == lhs.from && lhs.to == rhs.to
-	}
 }
 
 // Compares two arrays of TableSections and produces the operations


Sources/FunctionalTableData/TableSectionChangeSet.swift
```diff
@@ -11,10 +11,6 @@ import Foundation
 struct Moved<T: Equatable>: Equatable {
 	let from: T
 	let to: T
-
-	static func ==(lhs: Moved<T>, rhs: Moved<T>) -> Bool {
-		return lhs.from == lhs.from && lhs.to == rhs.to
```
raulriera: That `lhs.from == lhs.from`

raulriera: Unsure about a test for this 🤔 how come no test changed

raulriera:
Stop it bot

Sources/FunctionalTableData/TableSectionChangeSet.swift
dannbeau: I am wondering if it has anything to do with https://github.com/Shopify/store/issues/11412
We have been tracking down this issue that ultimately creates the inconsistent table data exception but the change set is correct. I have to deep dive on how this struct is used.

PR #193, opened 2020-03-19
Support context menus via cell actions

DrakeMorin:
This PR introduces support for Context Menus. Context Menus were released by Apple in iOS 13 and are intended to eventually replace 3D Touch interactions.

This PR adds a new CellAction type, ContextMenuConfiguration. Consumers use this struct to describe the context menu that should be used for a given cell. Importantly, ContextMenuConfiguration is not dependent on any iOS version. This means that consumers can specify the configuration without needing `@available(...)` annotations.

Upon further reflection, I have changed it so that devices running iOS 13+ will only have access to context menus. This is more in line with Apple's intent to deprecate Peek and Pop.
~~To enable support for context menus, consumers must pass `enableContextMenus: true` in the constructor for Functional Table Data. If the device is running iOS 13 or later, it will enable support for context menus. Please note that tableViews and collectionViews can only support Context Menus or 3D Touch. This is an intentional choice from Apple, and is reflected in this code as well.~~


Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UIContextMenuInteractionDelegate.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//  Created by Drake Morin on 2020-03-09.
+//  Copyright © 2020 Shopify. All rights reserved.
+//
+
+import UIKit
+
+extension FunctionalCollectionData {
+	@available(iOS 13.0, *)
+	class CollectionViewContextMenuDelegate: Delegate {
```
DrakeMorin: This separate delegate is needed because of how Apple has implemented Context Menus. Once a delegate implements the function `collectionView(_: contextMenuConfigurationForItemAt: point:)`, then 3D Touch interactions are automatically disabled. 

This means that if I just added these methods to the existing Delegate, I would irreversibly break every 3D Touch implementation on all devices running iOS 13+. Separating it into its own subclass avoids this entirely. It does require that FunctionalCollectionData be initialized with the correct delegate.

Sources/FunctionalTableData/CellActions.swift
```diff
@@ -173,7 +277,8 @@ public struct CellActions {
 		canPerformAction: CanPerformAction? = nil,
 		canBeMoved: Bool = false,
 		visibilityAction: VisibilityAction? = nil,
-		previewingViewControllerAction: PreviewingViewControllerAction? = nil) {
+		previewingViewControllerAction: PreviewingViewControllerAction? = nil,
+		contextMenuConfiguration: ContextMenuConfiguration? = nil) {
```
g-Off: Why not allow passing in an instance of `UIContextMenuConfiguration` directly instead of having the `ContextMenuConfiguration` that then needs to be mapped to that? It seems like we are mostly duplicating the class...

Sources/FunctionalTableData/CellActions.swift
```diff
@@ -117,6 +117,105 @@ public struct CellActions {
...
+		///   - menuActionsProvider: Closure that is executed when the system needs the actions to use in the context menu.
+		public init(title: String = "", preivewContentProvider: PreviewContentProvider? = nil, previewContentCommitter: PreviewContentCommitter? = nil, menuActionsProvider: @escaping MenuActionsProvider) {
+			self.title = title
+			self.previewContentProvider = preivewContentProvider
+			self.previewContentCommitter = previewContentCommitter
+			self.menuActionsProvider = menuActionsProvider
+		}
+		
+		@available(iOS 13.0, *)
+		internal func asUIContextMenuConfiguration(with identifier: IndexPath? = nil) -> UIContextMenuConfiguration {
```
g-Off: I think the `ItemPath` would be preferred over the `IndexPath` for a couple reasons:
- it would always be non-nil
- it should be relatively static, meaning that if an item moves from one index path to another (but is still the same thing) then it's ItemPath doesn't change (only if it moves sections)

Sources/FunctionalTableData/CellActions.swift
DrakeMorin: We are. The reason we don't allow for that is we can't store a `UIContextMenuConfiguration` in CellActions without making the entire struct only available to iOS 13+. I did look at storing it as an `AnyObject` but requiring all the extra casting felt wrong to me.

Sources/FunctionalTableData/CellActions.swift
DrakeMorin: This is great context I didn't have. Thank you!

Sources/FunctionalTableData/CellActions.swift
g-Off: I suggest we store it internally as an `Any?` and exposing it via a getter/setter with availability as it reduces the amount of code we have to manage and helps also keep us from having to maintain the mapping if/when Apple adds new properties to the `UIContextMenuConfiguration` type. Ideally we are able to stick as closely to vanilla UIKit as possible and only add the abstractions when absolutely necessary.

probably along the lines of:
```
private var _contextMenuConfiguration: Any?
@available(iOS 13.0, *)
public var contextMenuConfiguration: UIContextMenuConfiguration? {
    get { return _contextMenuConfiguration as! UIContextMenuConfiguration? }
    set { _contextMenuConfiguration = newValue }
}
```

We will also end up with an extra initializer. It would potentially need the ability to have `contextMenuConfiguration` passed in as an `Any?` in order to prevent having the call site worry about having two different init calls based on 13+ or <13. The init would be responsible for making sure that if the object is non-nil that it's type is correct.

Another option would be to leave it out of the `init` entirely and consumers would only be able to set via the property:
```
var actions = CellActions(...)
if #available(iOS 13, *) {
    actions.contextMenuConfiguration = UIContextMenuConfiguration(...)
}
```

Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UIContextMenuInteractionDelegate.swift
DrakeMorin: After additional exploration, I believe a better course of action is to only support context menus on iOS 13+. This is more in line with Apple's intention to deprecate Peek and Pop. It also negates the need for extra delegates, or modifying the `responds(to:)` function.

Sources/FunctionalTableData/CellActions.swift
DrakeMorin: I did explore this solution, but there is an issue with it. Specifically, that we make use of the `identifier` field on the `UIContextMenuConfiguration`.

In the delegate function: `tableView(_: contextMenuConfigurationForRowAt: point:)` we invoke `asUIContextMenuConfiguration(with:)` on the Context Menu Configuration abstraction struct. This builds a `UIContextMenuConfiguration` and sets the identifier appropriately. We use the identifier in the following delegate function to identify what cell a context menu belongs to:
```Swift
@available(iOS 13.0, *)
public func tableView(_ tableView: UITableView, willPerformPreviewActionForMenuWith configuration: UIContextMenuConfiguration, animator: UIContextMenuInteractionCommitAnimating) {
	guard let indexPath = configuration.identifier as? IndexPath else { return }
	let cellConfig = data.sections[indexPath]
	animator.addCompletion {
		cellConfig?.actions.contextMenuConfiguration?.previewContentCommitter?(animator.previewViewController)
	}
}
```

`identifier` is a get only property on `UIContextMenuConfiguration`. If we store the `UIContextMenuConfiguration` directly in the CellActions struct, we will not be able to change the identifier. We instead have to rely on the consumer having properly set the identifier. This seems more error prone than handling it ourselves.

I'm not sure how else we could solve this problem, but by keeping the `ContextMenuConfiguration` abstraction struct. I would love to hear your thoughts on the matter.

Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
```diff
@@ -113,5 +113,27 @@ extension FunctionalCollectionData {
 			
 			return proposedIndexPath
 		}
+		
+		@available(iOS 13.0, *)
+		public func collectionView(_ collectionView: UICollectionView, contextMenuConfigurationForItemAt indexPath: IndexPath, point: CGPoint) -> UIContextMenuConfiguration? {
+			let section = data.sections[indexPath.section]
+			let row = section.rows[indexPath.row]
```
nabilmaad, line 120: Could there ever be a case where any of these 2 could fail and throw an `Index out of range`?

Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
```diff
@@ -113,5 +113,27 @@ extension FunctionalCollectionData {
...
+			let cellConfig = data.sections[indexPath]
+			return cellConfig?.actions.contextMenuConfiguration?.asUIContextMenuConfiguration(with: ItemPathCopyable(itemPath: itemPath))
+		}
+		
+		@available(iOS 13.0, *)
+		public func collectionView(_ collectionView: UICollectionView, willPerformPreviewActionForMenuWith configuration: UIContextMenuConfiguration, animator: UIContextMenuInteractionCommitAnimating) {
+			guard let itemPathCopyable = configuration.identifier as? ItemPathCopyable else { return }
+			let keyPath = itemPathCopyable.itemPath
+			
+			guard let sectionIndex = data.sections.firstIndex(where: { $0.key == keyPath.sectionKey }), let rowIndex = data.sections[sectionIndex].rows.firstIndex(where: { $0.key == keyPath.itemKey })  else { return }
```
nabilmaad, line 131: Would be great to have each unbinding on its own line

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
```diff
@@ -227,5 +227,27 @@ extension FunctionalTableData {
 			let cellConfig = data.sections[indexPath]
 			return cellConfig?.actions.trailingActionConfiguration?.asSwipeActionsConfiguration()
 		}
+		
+		@available(iOS 13.0, *)
+		public func tableView(_ tableView: UITableView, contextMenuConfigurationForRowAt indexPath: IndexPath, point: CGPoint) -> UIContextMenuConfiguration? {
+			let section = data.sections[indexPath.section]
+			let row = section.rows[indexPath.row]
```
nabilmaad, line 234: Same question as for `UICollectionView`. Just pointing it out in case it's worth a check if those indices exist first as I don't have enough context to know if `data` will always contain items at the index paths invoked here

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
```diff
@@ -227,5 +227,27 @@ extension FunctionalTableData {
...
+			let cellConfig = data.sections[indexPath]
+			return cellConfig?.actions.contextMenuConfiguration?.asUIContextMenuConfiguration(with: ItemPathCopyable(itemPath: itemPath))
+		}
+		
+		@available(iOS 13.0, *)
+		public func tableView(_ tableView: UITableView, willPerformPreviewActionForMenuWith configuration: UIContextMenuConfiguration, animator: UIContextMenuInteractionCommitAnimating) {
+			guard let itemPathCopyable = configuration.identifier as? ItemPathCopyable else { return }
+			let keyPath = itemPathCopyable.itemPath
+
+			guard let sectionIndex = data.sections.firstIndex(where: { $0.key == keyPath.sectionKey }), let rowIndex = data.sections[sectionIndex].rows.firstIndex(where: { $0.key == keyPath.itemKey })  else { return }
```
nabilmaad, line 245: Same comment to break this down to help with readability

Sources/FunctionalTableData/CellActions.swift
```diff
@@ -117,6 +117,105 @@ public struct CellActions {
 			return configuration
 		}
 	}
+	
+	public struct ContextMenuConfiguration {
```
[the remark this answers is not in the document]
DrakeMorin: The `as...()` function is for internal use within FunctionalTableData. Specifically, the TableView and CollectionView delegates. By keeping the struct available to all iOS versions, it makes it easier for consumers to use it. They can configure the struct for all iOS versions without needing to deal with `@available` annotations, and leave it to us to apply it properly for iOS 13+.

Sources/FunctionalTableData/CellActions.swift
```diff
@@ -117,6 +117,105 @@ public struct CellActions {
...
+					style = .disabled
+				case .destructive:
+					style = .destructive
+				default:
+					style = UIMenuElement.Attributes()
+				}
+				
+				switch self.type {
+				case .action(let action):
+					return UIAction(title: title, image: image, identifier: nil, attributes: style, handler: { _ in action() } )
```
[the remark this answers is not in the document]
DrakeMorin: Good catch!

Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
DrakeMorin: That's better!

Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
DrakeMorin: It is unlikely, but possible that it could happen if we are generating a context menu while a render occurs. I've added a guard to avoid this.

Sources/FunctionalTableData/CellActions.swift
```diff
@@ -117,6 +117,105 @@ public struct CellActions {
...
+		let previewContentCommitter: PreviewContentCommitter?
+		let menuActionsProvider: MenuActionsProvider
+		
+		/// Creates a context menu configuration object with the specified set of actions.
+		///
+		/// - Parameters:
+		///   - title: The title of the context menu. Defaults to `""`.
+		///   - previewContentProvider: Closure that provides an custom preview view controller for the context menu. Defaults to `nil`.
+		///   - previewContentComitter: Closure that is executed when the user taps the context menu preview and should navigate to that screen. Defaults to `nil`.
+		///   - menuActionsProvider: Closure that is executed when the system needs the actions to use in the context menu.
```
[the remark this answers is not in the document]
DrakeMorin: Good catch

PR #194, opened 2020-05-12
adding basic accessibility struct

g-Off:
Starting point for a bit of extra accessibility data. The struct stores the iOS 13+ `userInputLabels` that are what the voice control system uses. This lets us put proper labels on cells so they can be tapped, instead of relying on the full accessibility string which can be quite cumbersome to speak.

### DIFF
--- a/Sources/FunctionalTableData/Accessibility.swift
+++ b/Sources/FunctionalTableData/Accessibility.swift
@@ -1,0 +1,42 @@
+//
+//  Accessibility.swift
+//  FunctionalTableData
+//
+//  Created by Geoffrey Foster on 2020-04-28.
+//  Copyright © 2020 Shopify. All rights reserved.
+//
+
+import UIKit
+
+public struct Accessibility: Equatable {
+	public var identifier: String?
+	public var userInputLabels: [String]?
+	
+	public init(identifier: String? = nil, userInputLabels: [String]? = nil) {
+		self.identifier = identifier
+		self.userInputLabels = userInputLabels
+	}
+	
+	internal func with(defaultIdentifier: String) -> Accessibility {
+		guard identifier == nil else { return self }
+		var copy = self
+		copy.identifier = defaultIdentifier
+		return copy
+	}
+}
+
+extension Accessibility {
+	func apply(to cell: UITableViewCell) {
+		cell.accessibilityIdentifier = identifier
+		if #available(iOS 13.0, *) {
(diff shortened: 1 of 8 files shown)


Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
```diff
@@ -31,8 +31,10 @@ extension FunctionalTableData {
 			let row = indexPath.row
 			let cellConfig = sectionData[row]
 			let cell = cellConfig.dequeueCell(from: tableView, at: indexPath)
-			cell.accessibilityIdentifier = ItemPath(sectionKey: sectionData.key, itemKey: cellConfig.key).description
-			
+			let accessibilityIdentifier = ItemPath(sectionKey: sectionData.key, itemKey: cellConfig.key).description
+			var accessibility = cellConfig.accessibility ?? Accessibility(identifier: accessibilityIdentifier)
```
alunbestor, line 35: Could we make `accessibility` default to `Accessibility()` rather than `nil`, to match what we do with `CellActions`? Effectively we're forcing each cell to have one anyway.

raulriera:
Missing tests 😉

g-Off:
> Unsure about the `nil` or default init. I assume we don't have that `ItemPath(sectionKey: sectionData.key, itemKey: cellConfig.key).description` information as a default implementation

Switched to a default `Accessibility()` that gets its identifier set to the `ItemPath` if and only if it's `nil` as per Alun's suggestion. Definitely cleaner this way.



> Missing tests 😉

🤫

Sources/FunctionalTableData/Accessibility.swift
```diff
@@ -0,0 +1,42 @@
...
+public struct Accessibility: Equatable {
+	public var identifier: String?
+	public var userInputLabels: [String]?
+	
+	public init(identifier: String? = nil, userInputLabels: [String]? = nil) {
+		self.identifier = identifier
+		self.userInputLabels = userInputLabels
+	}
+	
+	internal func with(identifier: String) -> Accessibility {
```
alunbestor, line 20: This feels a little misleading to me: UIKit's `with` methods (e.g. `UIFont.withSize`) and our own `with` methods (e.g. in Polaris.FontBook) always override the original value, whereas this specifies a fallback when the field doesn't already have a value.

Maybe just `with(defaultIdentifier:)`?

PR #197, opened 2020-05-27
Make reloadEntireTableThreshold configurable

gantonious:
If you have a table diff that changes more than 20 rows FTD will reload the entire table rather than applying all the changes separately. Unfortunately you lose animations when this happens. This makes the `reloadEntireTableThreshold` configurable at a global level to allow consumers to animate more than 20 changes at once.


gantonious:
Don't need this anymore. Ended up using DifferenceKit 🙂

PR #198, opened 2020-09-23
update readme

robinsenior:
- remove references to demo app.
- add SPM docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,6 @@ For more information, read our [documentation](https://shopify.github.io/Functio
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


raulriera (approved):
I assume you meant to delete the project file?

dannbeau (approved):

robinsenior:
@raulriera I didn't actually, good catch. I've reverted it, but also have no idea how that would've happened – I just checked my shell history and there's no reference to deleting it.

Was there a reason we deleted the demo project? My intern was hoping to learn from it.

raulriera:
It didn’t work with SPM if I remember correctly.  And it was also getting included 

But now that we can hide resources in the packages. We can bring it back and not include it

PR #199, opened 2020-10-21
Adds support for a selectionBackgroundProvider in the cellStyle.

desmondtsang:
The `CellStyle` already takes in a `BackgroundViewProvider` for the `backgroundView` property, but we can easily extend this idea to also take in a `BackgroundViewProvider` for the `selectedBackgroundView` property as well. 

This will allow for custom views on selection of cells.

### DIFF
--- a/Sources/FunctionalTableData/CellStyle.swift
+++ b/Sources/FunctionalTableData/CellStyle.swift
@@ -52,6 +52,8 @@ public struct CellStyle {
 	public var accessoryType: UITableViewCell.AccessoryType
 	/// The view's selection color.
 	public var selectionColor: UIColor?
+	/// Provide the view that is displayed behind the selected cell's other content.
+	public var selectionBackgroundViewProvider: BackgroundViewProvider?
 	/// The view's background color.
 	public var backgroundColor: UIColor?
 	/// The view that is displayed behind the cell's other content.
@@ -120,6 +122,7 @@ public struct CellStyle {
 				highlight: Bool? = nil,
 				accessoryType: UITableViewCell.AccessoryType = .none,
 				selectionColor: UIColor? = CellStyle.defaultSelectionColor,
+				selectionBackgroundViewProvider: BackgroundViewProvider? = nil,
 				backgroundColor: UIColor? = CellStyle.defaultBackgroundColor,
 				backgroundViewProvider: BackgroundViewProvider? = nil,
 				tintColor: UIColor? = nil,
@@ -133,6 +136,7 @@ public struct CellStyle {
 		self.highlight = highlight
 		self.accessoryType = accessoryType
 		self.selectionColor = selectionColor
+		self.selectionBackgroundViewProvider = selectionBackgroundViewProvider
 		self.backgroundColor = backgroundColor
 		self.backgroundViewProvider = backgroundViewProvider
 		self.tintColor = tintColor
@@ -163,6 +167,10 @@ public struct CellStyle {
 			cell.selectedBackgroundView = selectedBackgroundView
 		} else {
 			cell.selectedBackgroundView = nil
+		}
+		
+		if let selectedBackgroundView = selectionBackgroundViewProvider?.backgroundView() {
+			cell.selectedBackgroundView = selectedBackgroundView
 		}
 		
 		cell.layer.cornerRadius = cornerRadius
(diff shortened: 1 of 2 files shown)


jlock (approved):
Possibly this as well? (Vote cell)
image

Otherwise LGTM

Sources/FunctionalTableData/CellStyle.swift
```diff
@@ -159,6 +163,10 @@ public struct CellStyle {
 			cell.selectedBackgroundView = nil
 		}
 		
+		if let selectedBackgroundView = selectionBackgroundViewProvider?.backgroundView() {
+			cell.selectedBackgroundView = selectedBackgroundView
+		}
```
raulriera, line 168: Shouldn’t this code complete the color one? I should be able to set a custom view and color right? Either case, you should also undo anything you do in an else statement

desmondtsang: I wasn't sure if this entire solution is the correct way to go yet. 
Setting both a `selectedBackgroundColor` and `selectionColor` seems to override the `selectionColor`, similar to what happens if you provide both a `backgroundColor` and a `backgroundViewProvider`. 
🤔  Do you think it's more straightforward if we only respect the `selectedBackgroundViewProvider` if a `selectionColor` is not provided?

desmondtsang: Ah, so providing both a `backgroundColor` and a `backgroundSelectionViewProvider` sets the background colour and also the cell's `backgroundView`. In some circumstances, you won't see the resulting `backgroundColor`. 

But providing both a `selectionColor` and a `selectionBackgroundViewProvider` is different.. so maybe it should take an enum that can be either a `selectionColor` or a `selectionBackgroundViewProvider`, but not both.

raulriera: > Do you think it's more straightforward if we only respect the selectedBackgroundViewProvider if a selectionColor is not provided?

I would check what iOS is doing in this situation, what happens when you set a background view with a selection color?

But, as for my comment above `if let selectedBackgroundView = selectionBackgroundViewProvider?.backgroundView() {` the else statement of this should probably be removing the view, but that is going to clash with the previous selectionColor code. 

You might want to add tests to this new situation making sure the background and colors have the expected values depending on the configuration (color + background, color only, background only, etc)

desmondtsang: Setting the `selectionColor` overrides the `backgroundColor` for selected cells. 

Maybe it makes sense to do something like...
```
cell.selectedBackgroundView = nil
if let selectionColor = selectionColor {
	let selectedBackgroundView = UIView()
	selectedBackgroundView.backgroundColor = selectionColor
	cell.selectedBackgroundView = selectedBackgroundView
}
		
if let selectedBackgroundView = selectionBackgroundViewProvider?.backgroundView() {
	cell.selectedBackgroundView = selectedBackgroundView
}
```

raulriera (approved):
LGTM, thanks for adding the testa

Sources/FunctionalTableData/CellStyle.swift
```diff
@@ -151,12 +155,15 @@ public struct CellStyle {
...
 		if let selectionColor = selectionColor {
 			let selectedBackgroundView = UIView()
 			selectedBackgroundView.backgroundColor = selectionColor
 			cell.selectedBackgroundView = selectedBackgroundView
-		} else {
-			cell.selectedBackgroundView = nil
+		}
+		
+		if let selectedBackgroundView = selectionBackgroundViewProvider?.backgroundView() {
+			cell.selectedBackgroundView = selectedBackgroundView
```
stevethomp, line 166: nit: To be clear that you only want one, you could rearrange this to be:
```
if let selectedBackgroundView = selectionBackgroundViewProvider?.backgroundView() {
    cell.selectedBackgroundView = selectedBackgroundView
} else if let selectionColor = selectionColor {
    let selectedBackgroundView = UIView()
    selectedBackgroundView.backgroundColor = selectionColor
    cell.selectedBackgroundView = selectedBackgroundView
}
```
But I see why you followed the structure above, so I'm not strong on this.

stevethomp (approved):
👍  I like this change, makes sense and is consistent with the non selected backgrounds

PR #200, opened 2020-10-22
Support multi item selection via gestures

schmidyy:
[Apple docs](https://developer.apple.com/documentation/uikit/uitableviewdelegate/selecting_multiple_items_with_a_two-finger_pan_gesture)

---- 

This adds support for multi item selection in an FTD tableview. When the `tableView.allowsMultipleSelectionDuringEditing` is set to true, list multi selection can be initiated with a two finger swipe along the rows, or, if editing mode is already enabled, single finger swipe down the rows.

Lifecycle of this interaction is along the lines of:

- **`tableView(_:shouldBeginMultipleSelectionInteractionAt:) -> Bool`**: Determines if these gestures are available.
- **`tableView(_:didBeginMultipleSelectionInteractionAt:)`**: Called when a multi item selection gesture has begun. 
    - Handler forwarded to `CellActions` via `beginMultiSelectAction`
- **`tableView(_:didSelectRowAt:)`**: Called when a row is selected during the multi select gesture.
    - Handler forwarded to `CellActions` via `selectionAction`
- **`tableView(_:didDeselectRowAt:)`**: Called when a row is deselected during the multi select gesture.
    - Handler forwarded to `CellActions` via `deselectionAction`
- **`tableViewDidEndMultipleSelectionInteraction(_:)`**: Called when a multi item selection gesture has ended. 
    - ⚠️ This handler is **not forwarded to `CellActions`** because the API does not provide an `IndexPath`.

----

This is my first FTD PR. Let me know if I've missed anything here.

### DIFF
--- a/Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
+++ b/Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
@@ -255,5 +255,25 @@ extension FunctionalTableData {
 				cellConfig.actions.contextMenuConfiguration?.previewContentCommitter?(animator.previewViewController)
 			}
 		}
+		
+		/// Requires iOS 13.0+
+		/// Whether multi selection is available when editing mode is enabled.
+		/// Based on the tableView's `allowsMultipleSelectionDuringEditing` property.
+		/// https://developer.apple.com/documentation/uikit/uitableviewdelegate/selecting_multiple_items_with_a_two-finger_pan_gesture
+		@available(iOS 13.0, *)
+		public func tableView(_ tableView: UITableView, shouldBeginMultipleSelectionInteractionAt indexPath: IndexPath) -> Bool {
+			return tableView.allowsMultipleSelectionDuringEditing
+		}
+		
+		/// Requires iOS 13.0+
+		/// Called when a multi item selection gesture has begun.
+		/// - Handler forwarded to `CellActions` via `beginMultiSelectAction`
+		@available(iOS 13.0, *)
+		public func tableView(_ tableView: UITableView, didBeginMultipleSelectionInteractionAt indexPath: IndexPath) {
+			tableView.setEditing(true, animated: true)
+			
+			let cellConfig = data.sections[indexPath]
+			cellConfig?.actions.beginMultiSelectAction?()
+		}
 	}
 }
--- a/Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
+++ b/Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
@@ -258,11 +258,14 @@ extension FunctionalTableData {
 		
 		/// Requires iOS 13.0+
 		/// Whether multi selection is available when editing mode is enabled.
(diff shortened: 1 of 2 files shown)


antonino-u (approved):

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
```diff
@@ -255,5 +255,25 @@ extension FunctionalTableData {
...
 			}
 		}
+		
+		/// Requires iOS 13.0+
+		/// Whether multi selection is available when editing mode is enabled.
+		/// Based on the tableView's `allowsMultipleSelectionDuringEditing` property.
+		/// https://developer.apple.com/documentation/uikit/uitableviewdelegate/selecting_multiple_items_with_a_two-finger_pan_gesture
+		@available(iOS 13.0, *)
+		public func tableView(_ tableView: UITableView, shouldBeginMultipleSelectionInteractionAt indexPath: IndexPath) -> Bool {
+			return tableView.allowsMultipleSelectionDuringEditing
```
Caiopia, line 265: Any reason this check couldn't include the index paths as well to control which cells trigger multi select? 

```swift
public func tableView(_ tableView: UITableView, shouldBeginMultipleSelectionInteractionAt indexPath: IndexPath) -> Bool {
	let cellConfig = data.sections[indexPath]
	let shouldBeginMultiSelect = cellConfig?.actions.shouldBeginMultiSelectAction?() ?? false
	return tableView.allowsMultipleSelectionDuringEditing && shouldBeginMultiSelect
}
```

schmidyy: Could do that 👍  Not sure there's a use case where you'd only want multi select available from certain cells, but no harm in supporting it

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
```diff
@@ -255,5 +255,28 @@ extension FunctionalTableData {
...
+			let shouldBeginMultiSelect = cellConfig?.actions.shouldBeginMultiSelectAction?() ?? false
+			return tableView.allowsMultipleSelectionDuringEditing && shouldBeginMultiSelect
+		}
+		
+		/// Requires iOS 13.0+
+		/// Called when a multi item selection gesture has begun.
+		/// - Handler forwarded to `CellActions` via `beginMultiSelectAction`
+		@available(iOS 13.0, *)
+		public func tableView(_ tableView: UITableView, didBeginMultipleSelectionInteractionAt indexPath: IndexPath) {
+			tableView.setEditing(true, animated: true)
```
guillianbalisi, line 276: Is it required that the tableView enter edit mode whenever using multi select?

antonino-u: Yes! 
Also, in order for it to work, the function `public func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool` needs to return true, otherwise the tableView will not have the checkbox accessory view to the left of the cell. Which I'm planning on adding either our next PR :)

Caiopia (approved):

guillianbalisi (approved):

PR #201, opened 2020-10-26
Added cell selection state to cell actions.

antonino-u:
`UITableViewCell` has a selected property, which is automatically toggled through the UITableView in many cases. For example, if a user taps on a cell it is first highlighted then selected. If a user swipes on a cell, then the table view determines that that cell is now selected and shows the leading or trailing swipe actions accordingly.

If the table view supports multi-selection, then the table view shows checkboxes as a left accessory view of the cell, and if a cell is selected (via either a double-finger swipe gesture or a single tap), then its index path is added to `indexPathsForSelectedRows` automatically. 

If, however, the tableView is reloaded, or simply displaying for the first time and we want certain cells pre-selected, then `indexPathsForSelectedRows` is reset to nil. Therefore, it isn't a reliable way of keeping track of which rows are selected. `selected: Bool?` in `CellStyle` fixes this and now whenever the table view is rendered or its data is reloaded (pull-to-refresh, pagination, etc), the selections are not lost. 

In the same vein, just like how the table view needs to be in editing mode for multi-select to work, the function `public func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool` is called on `UITableViewDataSource` to confirm that an individual cell *can* have a checkbox in multi-select mode. Therefore, `selected` is checked in the internal `hasEditActions` computed property to make sure that the right cells can be selected and deselected.

### DIFF
--- a/Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
+++ b/Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
@@ -264,7 +264,7 @@ extension FunctionalTableData {
 		public func tableView(_ tableView: UITableView, shouldBeginMultipleSelectionInteractionAt indexPath: IndexPath) -> Bool {
 			let cellConfig = data.sections[indexPath]
 			
-			let shouldBeginMultiSelect = cellConfig?.actions.shouldBeginMultiSelectAction?() ?? false
+			let shouldBeginMultiSelect = cellConfig?.actions.shouldBeginMultiSelectGestureAction?() ?? false
 			return tableView.allowsMultipleSelectionDuringEditing && shouldBeginMultiSelect
 		}
 		
@@ -276,7 +276,7 @@ extension FunctionalTableData {
 			tableView.setEditing(true, animated: true)
 			
 			let cellConfig = data.sections[indexPath]
-			cellConfig?.actions.didBeginMultiSelectAction?()
+			cellConfig?.actions.didBeginMultiSelectGestureAction?()
 		}
 	}
 }
--- a/Sources/FunctionalTableData/CellActions.swift
+++ b/Sources/FunctionalTableData/CellActions.swift
@@ -227,8 +227,8 @@ public struct CellActions {
 	public typealias DeselectionAction = (_ sender: UIView) -> SelectionState
 	public typealias CanPerformAction = (_ selector: Selector) -> Bool
 	public typealias VisibilityAction = (_ cell: UIView, _ visible: Bool) -> Void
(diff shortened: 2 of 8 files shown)


Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDelegate.swift
```diff
@@ -152,6 +152,14 @@ extension FunctionalTableData {
 			}
 			
 			if let cellConfig = data.sections[indexPath] {
+				switch cellConfig.actions.selectionState {
+				case .selected:
+					tableView.selectRow(at: indexPath, animated: false, scrollPosition: .none)
+				case .deselected:
+					tableView.deselectRow(at: indexPath, animated: false)
```
schmidyy, line 159: Any reason we don't want the selection / deselection to be animated?

antonino-u: Yes! Since we're in `willDisplay cell`, the times when this is called and we actually want to select a cell will likely *not* be the first time this cell is presenting (ie reload, scrolling, paginating new cells into the table). I figure if someone wanted to select it the first time the cell shows up and animate it then can just have it show up as deselected and then select the cell by calling `tableView.selectRow(at indexPath, animated: true...)`

schmidyy (approved):
:chef-kiss:

raulriera:
Unsure if CellActions should be handling the SelectionState, but there are 2 legitimate things I think this PR should address:

- UICollectionView support
- Tests

antonino-u:
> Unsure if CellActions should be handling the SelectionState, but there are 2 legitimate things I think this PR should address:
> 
> * UICollectionView support
> * Tests

Hey @raulriera ! Take another look. I added a test, and moved selection to CellStyle and it works nicely! 
Note: the test failure has nothing to do with my changes:

`xcodebuild: error: Failed to build project FunctionalTableData with scheme FunctionalTableData-Package.
15
	Reason: Cannot test target “FunctionalTableDataTests” on “iPhone 8”: iPhone 8’s iOS Simulator 13.4.1 doesn’t match FunctionalTableDataTests’s iOS Simulator 14.0 deployment target.
16`

What should I do about this? Update github actions?

antonino-u:
@raulriera and I had an offline conversation where we discussed the right place for the selection state, and I ended up agreeing that CellStyle was probably better. Also, I have created an issue to apply this to `UICollectionView` and in the documentation for the property I mentioned that it was only for `UITableView`. I also added tests for it. I will merge this because it's going to make my life way easier, and if anyone would like to request changes I'm happy to discuss further :)

raulriera:
Thanks or the changes

PR #202, opened 2020-10-27
Unhighlight cell on deselection.

antonino-u:
Because of our custom selection logic for `UITableView`, cells that are selected are usually also highlighted using `CellStyler`.
The problem comes when gestures are used for multi-selection, or accessory view are tapped when in multi-selection, the last selected cell is always set as the highlighted cell. When using gestures or tapping the accessory view, the normal flow of `shouldHighlightCellAt` is not called, meaning that the cell does not properly un-highlight.
This PR simply adds logic to say: when a cell is de-selected, if it is still highlighted, unhighlight it.
In almost all cases except multi-select, this would have no effect because the cell is highlighted and un-highlighted before it is selected or de-selected.


schmidyy (approved):
LGTM!

antonino-u:
Given what I said [here](https://github.com/Shopify/FunctionalTableData/pull/201#issuecomment-719730869) and that this is a super simple PR and that @mats has approved it, I am merging this.

PR #204, opened 2020-10-29
Update test.yml

antonino-u:


.github/workflows/test.yml
```diff
@@ -15,6 +15,4 @@ jobs:
         run: swift package generate-xcodeproj
       - name: Testing
         run: |
-          xcodebuild clean test -project FunctionalTableData.xcodeproj -scheme FunctionalTableData-Package -destination "platform=iOS Simulator,OS=13.4.1,name=iPhone 8"
-        env:
-          DEVELOPER_DIR: /Applications/Xcode_11.4.1.app/Contents/Developer
+          xcodebuild clean test -project FunctionalTableData.xcodeproj -scheme FunctionalTableData-Package -destination "platform=iOS Simulator,name=iPhone 8"
```
schmidyy, line 18: Should we specify the OS version? What does it use if we don't?
```suggestion
          xcodebuild clean test -project FunctionalTableData.xcodeproj -scheme FunctionalTableData-Package -destination "platform=iOS Simulator,OS=14.0.0,name=iPhone 11"
```

antonino-u: It uses the sdk that matches whatever came with the current-default version of Xcode. 
I don't mind either way, I like being on the bleeding edge though :P

raulriera (approved):
Thanks for the fix!

PR #206, opened 2021-03-15
Use NSDiffableDataSource for iOS 13 or greater

jasonkemp9099:
To workaround crashes in UITableView with large change sets, use UITableViewDiffableDataSource when available (iOS 13 or greater)

Created an Impl protocol, with two implementations:

1. Original Functional Table Data algorithm
2. UITableViewDiffableDataSource-backed implementation for iOS 13 and up.

Then just followed the consequences of that. Any other changes are due to getting it working on IOS 13 at runtime. Tests with data exhibiting the crash with the original table data algorithm pass with diffable data sources.

### DIFF
--- a/Sources/FunctionalTableData/CellConfigType.swift
+++ b/Sources/FunctionalTableData/CellConfigType.swift
@@ -63,3 +63,82 @@ public extension CellConfigType {
 		return type(of: self) == type(of: other)
 	}
 }
+
+public struct AnyCellConfigType: CellConfigType, Hashable {
+
+	public static func ==(lhs: AnyCellConfigType, rhs: AnyCellConfigType) -> Bool {
+		return lhs.sectionKey == rhs.sectionKey && lhs.key == rhs.key 
+	}
+	
+	public var key: String {
+		return base.key
+	}
+	
+	public var style: CellStyle? {
+		get { return base.style }
+		set { base.style = newValue }
+	}
+	
+	public var actions: CellActions {
+		get { return base.actions }
+		set { base.actions = newValue }
+	}
+	
+	public var accessibility: Accessibility {
+		get { return base.accessibility }
+		set { base.accessibility = newValue }
+	}
+	
+	public var base: CellConfigType
+	private let sectionKey: String
+	
+	init(_ base: CellConfigType, sectionKey: String) {
+		self.base = base
+		self.sectionKey = sectionKey
+	}
+	
+	public init(_ base: CellConfigType) {
+		self.init(base, sectionKey: "")
+	}
+	
+	public func hash(into hasher: inout Hasher) {
+		hasher.combine(key)
+		hasher.combine(sectionKey)
+		hasher.combine(style)
+	}
+	
+	public func dequeueCell(from tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
+		return base.dequeueCell(from: tableView, at: indexPath)
+	}
+	
+	public func update(cell: UITableViewCell, in tableView: UITableView) {
+		base.update(cell: cell, in: tableView)
+	}
+	
+	public func update(cell: UICollectionViewCell, in collectionView: UICollectionView) {
+		base.update(cell: cell, in: collectionView)
+	}
+	
+	public func isEqual(_ other: CellConfigType) -> Bool {
+		guard let other = other as? AnyCellConfigType else { return false }
+		return sectionKey == other.sectionKey && key == other.key
+	}
+	
+	public func isSameKind(as other: CellConfigType) -> Bool {
+		return base.isSameKind(as: other)
+	}
+	
+	public func debugInfo() -> [String : Any] {
+		return base.debugInfo()
+	}
+	
+	public func register(with tableView: UITableView) {
+		base.register(with: tableView)
+	}
+	
+	public func register(with collectionView: UICollectionView) {
+		base.register(with: collectionView)
+	}
+	
+	
+}
--- a/Sources/FunctionalTableData/TableSectionHeaderFooter.swift
+++ b/Sources/FunctionalTableData/TableSectionHeaderFooter.swift
@@ -27,6 +27,10 @@ public protocol TableHeaderFooterStateType: Equatable {
 }
 
 public struct TableSectionHeaderFooter<ViewType: UIView, Layout: TableItemLayout, S: TableHeaderFooterStateType>: TableHeaderFooterConfigType {
+	public func dequeueCell(from tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
+		return UITableViewCell()
+	}
+	
 	public typealias ViewUpdater = (_ header: TableHeaderFooter<ViewType, Layout>, _ state: S) -> Void
 	public let state: S?
 	let updateView: ViewUpdater?
--- a/Sources/FunctionalTableData/TableView/FunctionalTableData+DiffableDataSource.swift
+++ b/Sources/FunctionalTableData/TableView/FunctionalTableData+DiffableDataSource.swift
@@ -1,0 +1,127 @@
+//
+//  FunctionalTableData+DiffableDataSource.swift
+//  FunctionalTableData
+//
+//  Created by Jason Kemp on 2021-03-30.
+//  Copyright © 2021 Shopify. All rights reserved.
+//
+
+import UIKit
+
+extension FunctionalTableData {
+	@available(iOS 13.0, *)
+	class DiffableDataSource: UITableViewDiffableDataSource<DiffableTableSection, AnyCellConfigType> {
+		let cellStyler: CellStyler
+		
+		var data: TableData {
+			return cellStyler.data
+		}
+		
+		init(tableView: UITableView, cellStyler: CellStyler) {
+			self.cellStyler = cellStyler
+			super.init(tableView: tableView) { (tableView, indexPath, cellConfigType) in
+				let sectionData = cellStyler.data.sections[indexPath.section]
+				let cell = cellConfigType.dequeueCell(from: tableView, at: indexPath)
+				let accessibilityIdentifier = ItemPath(sectionKey: sectionData.key, itemKey: cellConfigType.key).description
+				cellConfigType.accessibility.with(defaultIdentifier: accessibilityIdentifier).apply(to: cell)
+				cellStyler.update(cell: cell, cellConfig: cellConfigType, at: indexPath, in: tableView)
+				return cell
+			}
+		}
+		
+		override func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
+			guard let cellConfig = data.sections[indexPath] else { return false }
+			return cellConfig.actions.hasEditActions || self.tableView(tableView, canMoveRowAt: indexPath) || cellConfig.style?.selected != nil
+		}
+		
+		override func tableView(_ tableView: UITableView, canMoveRowAt indexPath: IndexPath) -> Bool {
+			return data.sections[indexPath]?.actions.canBeMoved ?? false
+		}
+		
+		override func tableView(_ tableView: UITableView, moveRowAt sourceIndexPath: IndexPath, to destinationIndexPath: IndexPath) {
+			// Should only ever be moving within section
+			assert(sourceIndexPath.section == destinationIndexPath.section)
+			
+			// Update internal state to match move
+			let cell = cellStyler.data.sections[sourceIndexPath.section].rows.remove(at: sourceIndexPath.row)
+			cellStyler.data.sections[destinationIndexPath.section].rows.insert(cell, at: destinationIndexPath.row)
+			cellStyler.data.sections[sourceIndexPath.section].didMoveRow?(sourceIndexPath.row, destinationIndexPath.row)
+			super.tableView(tableView, moveRowAt: sourceIndexPath, to: destinationIndexPath)
+		}
+	}
+	
+	@available(iOS 13.0, *)
+	class DiffableDataSourceFunctionalTableDataImpl: FunctionalTableDataImpl {
+		var datasource: DiffableDataSource!
+		var isRendering: Bool = false
+		let name: String
+		let cellStyler: CellStyler
+		
+		var tableView: UITableView? {
+			didSet {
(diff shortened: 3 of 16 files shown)


Sources/FunctionalTableData/TableSectionHeaderFooter.swift
```diff
@@ -27,6 +27,10 @@ public protocol TableHeaderFooterStateType: Equatable {
 }
 
 public struct TableSectionHeaderFooter<ViewType: UIView, Layout: TableItemLayout, S: TableHeaderFooterStateType>: TableHeaderFooterConfigType {
+	public func dequeueCell(from tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
+		return UITableViewCell()
+	}
```
Campbell-Scott, line 32: Why is this needed for `TableHeaderFooterConfigType`? `AnyCellConfigType` conformance?
Is there any consequence to returning an empty cell here - are we ever expecting this to be called on this type?

Would it make sense to have an `AnyTableHeaderConfigType` that leans on `func dequeueHeaderFooter(from tableView: UITableView) -> UITableViewHeaderFooterView?` instead?

jasonkemp9099: DIffable data sources fails with a crash because of a default implementation of `dequeueCell()` that returned `UITableViewCell.cellIdentifier`. Deleting that is in the diff here as well. I suspect it has to do with default implementations in protocol extensions and inheritance chains, sometimes those aren't predictable.

From what I can tell, `TableHeaderFooterConfigType` is also a `CellConfigType`, so that dequeueCell is necessary. I can't think of a scenario where we call that when using the TableHeaderFooterConfigType as a TableHeaderFooterConfigType, but if someone chooses to also use it as a CellConfigType and doesn't do the necessary work in their implementation, that could spell trouble. That seems remote enough as a possibility to do this.

josephroquedev:
NSDiffableDataSource is dropping cells if they have the same ID/content as a cell in a different section.

| With same ID / different content | With same ID / same content |
|---|---|
| Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 10 52 59 | Simulator Screen Shot - iPhone 12 Pro - 2021-03-17 at 10 53 13 |

Only the latter cell with identical content is being displayed. Also getting the following message in the console: `[UIDiffableDataSource] Warning: 2 inserted identifier(s) already present; existing items will be moved into place for this current insertion. Please note this will impact performance if items are not unique when inserted.`

If I change the content inside the cell or make the cell IDs globally unique, then they both render. Not an issue with classic diffing strategy

jasonkemp9099:
> NSDiffableDataSource is dropping cells if they have the same ID/content as a cell in a different section.

Nice find. Has to do with my Hashable implementations, I bet. I should make a snapshot test for this too.

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
```diff
@@ -57,4 +57,248 @@ extension FunctionalTableData {
...
+		
+		private func dumpDebugInfoForChanges(_ changes: TableSectionChangeSet, previousSections: [TableSection], visibleIndexPaths: [IndexPath], exceptionReason: String?, exceptionUserInfo: [AnyHashable: Any]?) {
+			guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
+			let exception = Exception(name: name, newSections: dataSource.data.sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: tableView?.frame ?? .zero, reason: exceptionReason, userInfo: exceptionUserInfo)
+			exceptionHandler.handle(exception: exception)
+		}
+	}
+	
+	@available(iOS 13.0, *)
+	class DiffableDataSourceFunctionalTableDataImpl: FunctionalTableDataImpl {
```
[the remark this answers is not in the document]
jasonkemp9099: Yes, duplicated the test specifically for perf tests and any others I could find that used a `FunctionalTableData` instance, now that we make it opt-in to diffable data sources.

Sources/FunctionalTableData/CellConfigType.swift
```diff
@@ -63,3 +63,73 @@ public extension CellConfigType {
...
+	private let sectionKey: String
+	
+	init(_ base: CellConfigType, sectionKey: String) {
+		self.base = base
+		self.sectionKey = sectionKey
+	}
+	
+	func hash(into hasher: inout Hasher) {
+		hasher.combine(key)
+		hasher.combine(accessibility)
```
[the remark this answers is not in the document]
jasonkemp9099: I'll pull that, I think. It came before adding `sectionKey`, the ol' kitchen sink approach. :)
I think using AnyCellConfigType to tie section and configType together for both hashable and equality testing is the trick to using diffable data sources here, after the first round of testing and doing some digging online.

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
```diff
@@ -57,4 +57,248 @@ extension FunctionalTableData {
...
+			tableView?.registerCellsForSections(localSections)
+			cellStyler.data.sections = localSections
+			
+			var snapshot = NSDiffableDataSourceSnapshot<TableSection, AnyCellConfigType>()
+			for newSection in localSections {
+				snapshot.appendSections([newSection])
+				snapshot.appendItems(newSection.anyRows)
+			}
+			
+			datasource?.apply(snapshot, animatingDifferences: false, completion: completion)
```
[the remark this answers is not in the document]
jasonkemp9099: OK, figured this out. 

UITableViewDiffableDataSource offers `defaultRowAnimation` to customize the animation when a snapshot is applied. We can't customize down to each operation like previously. We pick `.fade` for insert and delete, and `.automatic` for reload. Tried both. Automatic is the default for any UITableView. It's much more prevalent using diffable data sources to my eyes. `.fade` is a bit nicer.

I'll set `defaultRowAnimation` to `animations.rows.insert` here for now.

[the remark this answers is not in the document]
jasonkemp9099: Another follow up! 😄

Scott posted some pretty compelling evidence in the issue channel in the form of videos to preferring `.none` as the default row animation

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
```diff
@@ -57,4 +57,248 @@ extension FunctionalTableData {
...
+			guard let exceptionHandler = FunctionalTableData.exceptionHandler else { return }
+			let exception = Exception(name: name, newSections: dataSource.data.sections, oldSections: previousSections, changes: changes, visible: visibleIndexPaths, viewFrame: tableView?.frame ?? .zero, reason: exceptionReason, userInfo: exceptionUserInfo)
+			exceptionHandler.handle(exception: exception)
+		}
+	}
+	
+	@available(iOS 13.0, *)
+	class DiffableDataSourceFunctionalTableDataImpl: FunctionalTableDataImpl {
+		var datasource: DiffableDataSource?
+		var isRendering: Bool = false
```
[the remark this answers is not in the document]
josephroquedev: Should we still make it true/false at the start/end of `renderAndDiff`, for posterity?

Sources/FunctionalTableData/CellConfigType.swift
```diff
@@ -63,3 +63,72 @@ public extension CellConfigType {
...
+		get { return base.actions }
+		set { base.actions = newValue }
+	}
+	
+	var accessibility: Accessibility {
+		get { return base.accessibility }
+		set { base.accessibility = newValue }
+	}
+	
+	private var base: CellConfigType
```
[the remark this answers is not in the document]
jasonkemp9099: Can't be `let` because CellConfigType is not constrained to objects and there are setters. Will make public.

Sources/FunctionalTableData/CellConfigType.swift
```diff
@@ -63,3 +63,72 @@ public extension CellConfigType {
...
+	
+	func update(cell: UITableViewCell, in tableView: UITableView) {
+		base.update(cell: cell, in: tableView)
+	}
+	
+	func update(cell: UICollectionViewCell, in collectionView: UICollectionView) {
+		base.update(cell: cell, in: collectionView)
+	}
+	
+	func isEqual(_ other: CellConfigType) -> Bool {
```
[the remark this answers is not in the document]
jasonkemp9099: Yeah, probably. Good call.

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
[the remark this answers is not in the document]
jasonkemp9099: Fine. But I'm only doing it for approvals! :)

Campbell-Scott (approved):
Code & test updates looking good ✅ 
Manual testing also looking good ✅

josephroquedev (approved):

Sources/FunctionalTableData/CellConfigType.swift
```diff
@@ -63,3 +63,80 @@ public extension CellConfigType {
...
 	}
 }
+
+public struct AnyCellConfigType: CellConfigType, Hashable {
+	public func dequeueCell(from tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
+		return base.dequeueCell(from: tableView, at: indexPath)
+	}
+	
+	public static func ==(lhs: AnyCellConfigType, rhs: AnyCellConfigType) -> Bool {
+		return lhs.sectionKey == rhs.sectionKey && lhs.key == rhs.key && lhs.base.isSameKind(as: rhs.base) && lhs.base.isEqual(rhs.base)
```
Caiopia, line 73: Any reason we can't we defer to the `isEqual` method here? I'd expect both `==` and `isEqual` methods to be equivalent.

```suggestion
		return lhs.isEqual(rhs)
```

jasonkemp9099: In this case, no.

Equality and hash have to use the same fields, otherwise the diffable data source will sometimes fail to apply the snapshot (with an exception).

We include the sectionKey here to tie each cell with its section to account for a cell with the same key and state in multiple sections. If we don't, diffable data sources just uses the last one it saw with the same hash. I think it stores sections and rows together in the same hashtable or something.

`AnyCellConfigType` is solely around for being Hashable, to satisfy the compiler and the requirements of diffable data sources. It should probably be internal, but there's no harm in making it public. For its `CellConfigType` methods, it should defer to its `base` instance.

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
```diff
@@ -57,4 +57,347 @@ extension FunctionalTableData {
 			return cellConfig.actions.hasEditActions || self.tableView(tableView, canMoveRowAt: indexPath) || cellConfig.style?.selected != nil
 		}
 	}
+	
+	class OldFunctionalTableDataImpl: FunctionalTableDataImpl {
```
Caiopia, line 61: Styling: Can we move some of these types to their own files to short this file up a bit?

Sources/FunctionalTableData/TableView/FunctionalTableData+UITableViewDataSource.swift
```diff
@@ -57,4 +57,347 @@ extension FunctionalTableData {
...
+				dataSource.defaultRowAnimation = .none
+				self.datasource = dataSource
+				
+			}
+		}
+		
+		init(name: String, cellStyler: CellStyler) {
+			self.name = name
+			self.cellStyler = cellStyler
+			self.applyQueue = DispatchQueue.global(qos: .background)
```
Caiopia, line 321: I imagine that snapshot applying is frequently tied to the user's interactions. Is there another queue with higher priority we could use? Maybe `.userInteractive`?

jasonkemp9099: Deleted this. Wasn't using the queue because it broke tests and didn't notice improvements.

Tests/FunctionalTableDataTests/DiffableDataSourceFunctionalTableDataPerformanceTests.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  FunctionaltableDataPerformanceTests.swift
```
Caiopia, line 2: Small detail:

```suggestion
//  DiffableDataSourceFunctionalTableDataPerformanceTests.swift
```

Sources/FunctionalTableData/TableView/FunctionalTableData+DiffableDataSource.swift
```diff
@@ -0,0 +1,127 @@
...
+			let changeSet = TableSectionChangeSet(old: oldSections, new: localSections, visibleIndexPaths: indexPaths)
+			datasource.data.sections = localSections
+			
+			var snapshot = NSDiffableDataSourceSnapshot<DiffableTableSection, AnyCellConfigType>()
+			let diffableSections = localSections.map { DiffableTableSection($0) }
+			snapshot.appendSections(diffableSections)
+			for newSection in diffableSections {
+				snapshot.appendItems(newSection.anyRows, toSection: newSection)
+			}
+			var reallyAnimated: Bool = true
```
Caiopia, line 90: nit: Haha, these variable names are great, but maybe something like `isFirstRender` or `isFirstSnapshot` could be a bit more descriptive, followed by `let shouldAnimate = animated && !isFirstRender`?

Caiopia (approved):
LGTM - Thanks for your patience with this and awesome work! I'm excited for this.

For visibility - There is a known limitation right now with the new diffing strategy where it can't detect certain state changes that result in a cell reload. This is due to the hashable implementation being mostly key based. A current temporary workaround is to embed some state into the key so that diffable data source can detect a state change and reload. However, long-term, a `Hashable` requirement for view states should eventually be added so that changes can be detected regardless of key.

PR #213, opened 2023-02-27
Prevent code from hitting IndexPath.section count trap

mauriciomeirelles:
We currently have a crash happening in prod on Shopify Mobile when an user is trying to move a media around in `ProductMediaViewController`. Not sure exactly what is happening still, but for some reason `collectionView(_ collectionView: UICollectionView, targetIndexPathForMoveFromItemAt originalIndexPath: IndexPath, toProposedIndexPath proposedIndexPath: IndexPath) -> IndexPath` is being called with one (or both) of the `IndexPath` with less/more than 2 items inside. 

Thing is `IndexPath.section.getter` implementation in UIKit has a trap inside for when it has less or more than 2 items. This is why we are having this crash when trying to access the section getter. So to prevent this crash from happening I'm adding an extra check for both `IndexPath` to make sure they have exactly 2 items, otherwise I return the `originalIndexPath` without crashing it. 

This should prevent the crash from happening in this instance.

Crash Log:
https://app.bugsnag.com/shopify/mobile-shopify-ios/errors/6386c2ccaacb9000086308ac?event_id=63f8d6fa00b811867ab80000&i=gh&m=ci

Mentions https://github.com/Shopify/mobile/issues/14974

### DIFF
--- a/Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
+++ b/Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
@@ -103,7 +103,7 @@ extension FunctionalCollectionData {
 		}
 		
 		public func collectionView(_ collectionView: UICollectionView, targetIndexPathForMoveFromItemAt originalIndexPath: IndexPath, toProposedIndexPath proposedIndexPath: IndexPath) -> IndexPath {
-			guard originalIndexPath.section == proposedIndexPath.section else {
+			guard originalIndexPath.count == 2, proposedIndexPath.count == 2, originalIndexPath.section == proposedIndexPath.section else {
 				return originalIndexPath
 			}
 			
--- a/FunctionalTableData.xcodeproj/project.pbxproj
+++ b/FunctionalTableData.xcodeproj/project.pbxproj
@@ -7,15 +7,28 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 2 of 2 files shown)


Sources/FunctionalTableData/CollectionView/FunctionalCollectionData+UICollectionViewDelegate.swift
```diff
@@ -103,7 +103,7 @@ extension FunctionalCollectionData {
 		}
 		
 		public func collectionView(_ collectionView: UICollectionView, targetIndexPathForMoveFromItemAt originalIndexPath: IndexPath, toProposedIndexPath proposedIndexPath: IndexPath) -> IndexPath {
-			guard originalIndexPath.section == proposedIndexPath.section else {
+			guard originalIndexPath.count == 2, proposedIndexPath.count == 2, originalIndexPath.section == proposedIndexPath.section else {
```
adam-felix-shopify, line 106: Curiosity question:  Where's the 2 coming from?

Caiopia:
> Btw the changes on the Xcode proj were because those files were not added to it, so the project was not even building. Not sure why it was like this but the only thing I did was to add them to the Xcode proj again, the files were already in the project folder even.

I bet that's because the project was made to be a Swift package. I wonder if we should just delete the Xcode project and update the `Manual` installation option to add a local swift package instead?

Caiopia:
> Thing is IndexPath.section.getter implementation in UIKit has a trap inside for when it has less or more than 2 items

Noticed this getter was mentioned in the stacktraces too... Found the answer to [this post](https://developer.apple.com/forums/thread/663787) which confirms what you're saying. Any ideas why the index path could have a length other than 2? Trying to understand why we might run into this in the first place.

